PR #7, opened 2024-01-23
FE-526 Feature/permission request settings

PavlosTze:
## **Why?**

Request for notification permission and create 2 new sections in preferences for permissions & announcements.

### **How?**
1. Added the "Announcements" entry via XML in the in-app settings and the respective on click listener on `PreferenceFragment`
2. Request on Android 13 devices and above notifications permissions (`POST_NOTIFICATIONS`). On Android 12 and below the permission is automatically asked when the first notification channel is being created.
3. Added the respective entry on the Manifest file.
4. Added the "Notifications" entry via XML in the in-app settings and the respective on listeners on `PreferenceFragment`, check its status at `onResume`.

### **Testing**
1. You will get 2 notification permission prompts, so reject the first one
5. Go to in-app settings and ensure that the notifications setting is OFF
6. Close and reopen the app
7. On the second notification prompt by the system accept the permission
2. Go to in-app settings and ensure that the notifications setting is ON
8. Check whether clicking the notifications setting goes to the app settings in your device and toggling the permission there toggles the respective in-app setting as well
9. Ensure that the "Announcements" section in Settings is OK.

### DIFF
--- a/app/src/main/java/com/weatherxm/ui/preferences/PreferenceFragment.kt
+++ b/app/src/main/java/com/weatherxm/ui/preferences/PreferenceFragment.kt
@@ -103,20 +103,27 @@ class PreferenceFragment : PreferenceFragmentCompat() {
             true
         }
 
-        analyticsPreference?.setOnPreferenceChangeListener { _, newValue ->
-            model.setAnalyticsEnabled(newValue as Boolean)
-            true
-        }
-
         model.isLoggedIn().observe(this) { result ->
             result
                 .mapLeft {
                     Timber.d("Not logged in. Hide account preferences.")
-                    onLoggedOut(shortWxmSurvey, logoutBtn, resetPassBtn, deleteAccountButton)
+                    onLoggedOut(
+                        shortWxmSurvey,
+                        logoutBtn,
+                        resetPassBtn,
+                        deleteAccountButton,
+                        analyticsPreference
+                    )
                 }
                 .map {
                     Timber.d("Logged in. Handle button clicks")
-                    onLoggedIn(shortWxmSurvey, logoutBtn, resetPassBtn, deleteAccountButton)
+                    onLoggedIn(
+                        shortWxmSurvey,
+                        logoutBtn,
+                        resetPassBtn,
+                        deleteAccountButton,
+                        analyticsPreference
+                    )
                 }
         }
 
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/weatherxm/ui/components/BaseInterface.kt
```diff
@@ -30,4 +34,17 @@ interface BaseInterface {
         }
         snackbar?.show()
     }
+
+    fun requestLocationPermissions(activity: FragmentActivity?, onGranted: () -> Unit) {
```
JohnFrontzos, line 38: nice!

JohnFrontzos:
We may should start to replace all the Serializable use with Parcelable. It's not in the scope of this PR but in ClaimLocationFragment, we could replace the
arguments = Bundle().apply { putSerializable(ARG_DEVICE_TYPE, deviceType) } with arguments = bundleOf(ARG_DEVICE_TYPE to deviceType) as the DeviceType is a Parcelable object already.

app/src/main/java/com/weatherxm/ui/home/HomeActivity.kt
```diff
@@ -34,6 +37,14 @@ class HomeActivity : BaseActivity(), BaseMapFragment.OnMapDebugInfoListener {
     private lateinit var binding: ActivityHomeBinding
     private lateinit var navController: NavController
 
+    init {
```
JohnFrontzos, line 40: Why we are using init & withCreate here and not move the permission's request inside onCreate() ?

JohnFrontzos (changes requested):

pantelisss (changes requested):
The allow notifications option should be visible when user is logged out

app/src/main/java/com/weatherxm/ui/home/HomeActivity.kt
PavlosTze: I thought it was cleaner and also in order to do it in a new Coroutine and not on the main thread (e.g. block the main thread), is it wrong or am I missing something?

app/src/main/java/com/weatherxm/ui/preferences/PreferenceFragment.kt
```diff
@@ -139,4 +166,36 @@ class PreferenceFragment : PreferenceFragmentCompat() {
             .build()
             .show(this)
     }
+
+    private fun handleNotificationsPreference() {
+        val notificationsPreference =
+            findPreference<SwitchPreferenceCompat>(getString(R.string.notifications_preference_key))
+
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.TIRAMISU) {
```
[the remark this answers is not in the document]
PavlosTze: Right on... Android Studio didn't suggested it, weird..

PavlosTze:
Good catches @pantelisss @JohnFrontzos , please check again :)

pantelisss:
@PavlosTze Analytics switch should not be visible for logged out users

app/src/main/java/com/weatherxm/ui/home/HomeActivity.kt
JohnFrontzos: It's not wrong, I was just curious about the thought behind the decision. 

I could see scope error when using class init, in Activity or Fragment lifecycle but in this specific case, we may not have any. You could also "not block" the main thread, by using another coroutine scope within the onCreate(). I don't have a strong opinion on this topic, so you could try whatever you think is best for this case.

JohnFrontzos (approved):

pantelisss (approved):
🤘🚀

PR #8, opened 2024-01-26
FE-528 Integration of Notifications

PavlosTze:
## **Why?**

Integrate notifications in order to be able to show them to users.

### **How?**

1. Created `WXMRemoteMessage` which will be our `data class` for these messages
2. Created `enum class RemoteMessageType(val id: String, val publicName: String, val desc: String)` where `id` is the respective `type` custom value as agreed, `publicName` & `desc` are texts that will be visible in app-settings for informing the user what this channel is about.
3. Proper implementation of `onMessageReceived` in order to handle notifications when the app is on foreground
4. When the app is in the background, and the notification gets clicked, `StartupActivity` starts with notification-data in its `intent` so we handle it there and if the app opened due to a notification we open the `UrlRouterActivity`.
5. `UrlRouterActivity` makes all the necessary checks as to what to do next and if a valid `url` is present it opens a Chrome Tab with `navigator.openWebsite(...)`. Also, we make a check `isTaskRoot` in order to first reopen the `StartupActivity` in case a user hits the "back button" to keep the app open after the ChromeTab gets closed (the `StartupActivity` handles properly as to "where" the user should go next).

### **Testing**

1. Get your FCM token by searching in Logcat "token"
7. Try sending a test message through Firebase Console
8. Test the app in different scenarios how it handles it, visible, on the background, stopped.

### DIFF
--- a/app/src/main/java/com/weatherxm/data/DataModels.kt
+++ b/app/src/main/java/com/weatherxm/data/DataModels.kt
@@ -7,6 +7,7 @@ import androidx.annotation.Keep
 import com.squareup.moshi.Json
 import com.squareup.moshi.JsonClass
 import kotlinx.parcelize.Parcelize
+import timber.log.Timber
 
 @Keep
 @JsonClass(generateAdapter = true)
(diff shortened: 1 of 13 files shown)


pantelisss (approved):
👌

PR #9, opened 2024-01-29
FE-533 Replace serializable with parcelable

PavlosTze:
## **Why?**

Replace Serializable with Parcelable.

### **How?**
1. Make `WidgetType` enum class a `Parcelable`.
2. Replace all get/put serializable calls to the parcelable ones
3. Replaced `BigInteger` in `cumulativeAmount`, `available` and `totalClaimed` to `Double` as it works fine with that and `BigInteger` increased the code complexity.

### **Testing**

Do a claiming through the Profile screen.


JohnFrontzos (approved):

pantelisss (approved):
As far as I can understand seems good

PR #11, opened 2024-01-30
FE-546 FE-542 Feature/bug fixes

PavlosTze:
## **Why?**
Fixed two minor bugs affecting the station's friendly name flow and the follow & historical charts flow.

### **How?**
1. Update the `CurrentViewModel`'s `device` when an updated one comes from the parent.
2. Update device's friendly name on `DeviceSettingsViewModel` when a success on editing or clearing the friendly name takes place.

### **Testing**
1. Make sure the bug as presented [here in the video](https://linear.app/weatherxm/issue/FE-542/research-historical-bugs-reported-by-george) is not reproducable anymore.
3. Make sure the bug as presented [here in the description](https://linear.app/weatherxm/issue/FE-546/cannot-rename-a-station-twice) is not reproducable anymore


app/src/main/java/com/weatherxm/ui/devicesettings/DeviceSettingsViewModel.kt
```diff
@@ -58,6 +58,7 @@ class DeviceSettingsViewModel(
                             Analytics.ParamValue.CHANGE_STATION_NAME_RESULT_ID.paramValue,
                             success = 1L
                         )
+                        device.friendlyName = friendlyName
                         onEditNameChange.postValue(friendlyName)
```
pantelisss, line 62: I don't know if is easy, It would be better to use the API response to update the `friendlyName`. In any case, it's minor

pantelisss (approved):
Just check my comment, it's minor

app/src/main/java/com/weatherxm/ui/devicesettings/DeviceSettingsViewModel.kt
PavlosTze: The API response doesn't contain the `friendlyName`, it's an empty body actually. 😞

PR #12, opened 2024-01-31
FE-555 Create a workflow_dispatch aka manual workflow to trigger a QA release

PavlosTze:
## **Why?**

We need a way for the QA team to manually trigger releases whenever they want.

### **How?**

- Followed the guide [here](https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/) and added a `workflow_dispatch` GitHub Workflow, which takes as input the environment parameters (`production` or `development`), sets the respective environment and creates the respective release.

### **Testing**

Not possible, based on [this](https://stackoverflow.com/a/74722590/5403137), the action needs to be on the `main` branch,

### DIFF
--- a/.github/workflows/on-qa-firebase-distribution.yml
+++ b/.github/workflows/on-qa-firebase-distribution.yml
@@ -1,7 +1,12 @@
-name: Build Production & Development and Distribute on Firebase
+name: Build QA Release and Distribute on Firebase
 on:
-    push:
-        branches: [ main ]
+    workflow_dispatch:
+        inputs:
+            buildType:
+                type: environment
+                description: 'Select between "production" and "development" as build types.'
+                required: true
+                default: 'production'
 env:
     # Claim DApp URL
     ORG_GRADLE_PROJECT_CLAIM_APP_URL: ${{ secrets.CLAIM_APP_URL }}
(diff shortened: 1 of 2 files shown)


.github/workflows/on-qa-firebase-distribution.yml
```diff
@@ -0,0 +1,64 @@
+name: Build QA Release and Distribute on Firebase
+on:
+    workflow_dispatch:
+        inputs:
+            buildType:
```
JohnFrontzos, line 5: Am not sure that this would work.  I would try something like 
`    inputs:
        environment:
          type: choice
          description: 'Select environment'
          required: true
          options:
            - Playground
            - Dev
            - Production
`  
Also, if needed,  we could add another option for the Release or Debug flavor.

PavlosTze: There is a specific `environment` type, not sure how it works though I've read it on documentation:
https://github.blog/changelog/2021-11-10-github-actions-input-types-for-manual-workflows/

JohnFrontzos: Would not have any problem but feel free to change it in the previous value.

PR #32, opened 2024-02-12
FE-580 Feature/annotation group

PavlosTze:
## **Why?**
Integrate new `annotation_summary` property.

### **How?**
- Added `annotationSummary` in the API Models and `RewardsAnnotationGroup` which is the new structure we will use.
- Sort the new summary annotations based on severity (error -> warning -> info) before populating the UI
- Minor enhancements in `MessageCardView` in order to implement the new annotations
- Changes in `RewardProblemsAdapter` to use the new `RewardsAnnotatonGroup` as input data

### **Testing**
Either run `localMockDebug` flavor or the dev flavor (which communicates with the dev API) and ensure everything is OK rewards-related.

### Others:
Replace the following:
1. `device.relation == DeviceRelation.OWNED`
2. `device.relation == DeviceRelation.FOLLOWED`
3. `device.relation == DeviceRelation.UNFOLLOWED`

With the respective extension functions:
1. `device.isOwned()`
2. `device.isFollowed()`
3. `device.isUnfollowed()`

### DIFF
--- a/app/src/main/java/com/weatherxm/ui/common/UIModels.kt
+++ b/app/src/main/java/com/weatherxm/ui/common/UIModels.kt
@@ -11,6 +11,7 @@ import com.weatherxm.data.Hex
 import com.weatherxm.data.HourlyWeather
 import com.weatherxm.data.Location
 import com.weatherxm.data.QoDErrorAffects
+import com.weatherxm.data.RewardsAnnotationGroup
 import com.weatherxm.data.RewardsAnnotations
 import com.weatherxm.data.RewardsObject
 import com.weatherxm.data.Transaction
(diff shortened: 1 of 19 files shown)


app/src/main/java/com/weatherxm/ui/common/UIModels.kt
```diff
@@ -453,3 +467,16 @@ enum class AnnotationCode : Parcelable {
...
+enum class AnnotationGroupCode : Parcelable {
+    NO_WALLET,
+    NO_STATION_DATA,
+    SENSOR_PROBLEMS,
+    WEATHER_DATA_GAPS,
+    BAD_STATION_DEPLOYMENT,
+    NO_LOCATION_DATA,
+    LOCATION_NOT_VERIFIED,
+    USER_RELOCATION_PENALTY,
+    UNKNOWN
```
pantelisss, line 481: I'm not sure if the client should know about the other cases since the logic is to make the client agnostic of annotations. 
IMO we should handle only the `NO_WALLET ` and `LOCATION_NOT_VERIFIED` cases

pantelisss (approved):
👌
Check my comment and is up to you

JohnFrontzos (approved):
LGTM! Tho the majority of changes out of this PR scope (talking about the helper functions)

app/src/main/java/com/weatherxm/ui/common/UIModels.kt
JohnFrontzos: I'll agree with @pantelisss on this. I can see a potential use of the others in the future, but it's not part of this feature and they could change names again.

PR #34, opened 2024-02-13
FE-486 Feature/upgrade kable optimize scopes

PavlosTze:
## **Why?**
Upgrade kable library and scope-related optimizations.

### **How?**
- Upgraded Kable library to 0.29.0
- Replaced `GlobalScope` invocations with other scopes where it's needed
- Minor fixes when closing the activity on Change Frequency & Reboot.
- Fix on retrying on Helium claiming when a BLE error is present

### **Testing**
Test all BLE-related functionalities, try with different combinations (paired vs non-paired):
- Claiming of Helium (we want BLE-related to be OK, and get an error at the end that device is already claimed)
- Reboot of Helium Station (through station settings)
- Device OTA of Helium Station **(if needed, ping me to send you a test version with the "Needs Update" flag enabled)**
- Change Frequency of Helium Station (through station settings)


pantelisss (approved):
Codewise looks good!

PR #38, opened 2024-02-20
FE-619 Create new data class for the updated rewards response

PavlosTze:
## **Why?**
The new API models needed for the updated rewards response.

### **How?**
Created `RewardsResponse` , `Reward` and `RewardsTimelineEntry` data classes.

### **Testing**
Ensure that the models match the spec's one.


pantelisss (approved):
LGTM 👌

PR #39, opened 2024-02-20
FE-607 FE-620 Redesign rewards tab

PavlosTze:
## **Why?**
A redesign of the rewards tab to better present the available information.

### **How?**
- Created `isMainnetEnabled` and `getMainnetMessage` functions to get the respective Remote Config values in order to check if we should present the mainnet banner and secondly to populate it with the correct message.
- Very minor fixes on reward details screen to build successfully until the next redesign which is coming up for this screen.
- Cleaned up some not used anymore `UIModels`
- Replaced `RewardsCardView` with the `DailyRewardsCardView` which is a reusable component of that new card.
- Changes in `RewardsFragment` in order to properly show the new UI design. Some pieces of the weekly streak card have being created with Jetpack compose.
- Created a new helper extension function `DayOfWeek.getFirstLetter` that we use in the Weekly Streak bars.
- Redesign of the respective XML files.

### **Testing**
- Run Local Mock flavor and play with the respective JSON `get_user_device_rewards.json` and ensure everything works properly in the rewards tab.
- Or run a build looking at the Dev API where the updated endpoint is deployed.


app/src/main/java/com/weatherxm/data/repository/AppConfigRepository.kt
```diff
@@ -13,6 +13,8 @@ interface AppConfigRepository {
     fun setLastRemindedVersion()
     fun getInstallationId(): String?
     fun getRewardsHideAnnotationThreshold(): Long
+    fun isMainnetEnabled(): Boolean
```
JohnFrontzos, line 16: Also, we could use a Pair<Boolean, String> fun result here, for the featureEnabled and the message variable.

JohnFrontzos (approved):
Love to see UI Compose! gg Pavlos

app/src/main/java/com/weatherxm/data/repository/AppConfigRepository.kt
PavlosTze: I agree, but we may need `isMainnetEnabled` for other changes as well (e.g. the text changes we had discussed when we get on mainnet) so that's why I created a specific, short-lived function.

PR #45, opened 2024-02-27
FE-657 Feature/reward issues screen

PavlosTze:
## **Why?**
Implementation of the new Reward Issues screen.

### **How?**
- Created a new `RewardIssuesActivity` which takes as arguments the `Reward` object and the `UIDevice` object.
- This activity uses a respective XML file with a recycler view, using the `RewardIssuesAdapter` to populate the UI.

### **Testing**
**For testing purposes only, on the reward tab the "View Reward Details" opens the Reward Issues screen.**

- Either run local mock flavor
- Or in your dev build find a station with >= 1 issues to view the new reward issues screen.


app/src/main/java/com/weatherxm/ui/components/HeaderView.kt
```diff
@@ -0,0 +1,53 @@
...
+
+import android.content.Context
+import android.util.AttributeSet
+import android.view.LayoutInflater
+import android.widget.LinearLayout
+import com.weatherxm.R
+import com.weatherxm.databinding.ViewHeaderBinding
+import org.koin.core.component.KoinComponent
+
+open class HeaderView : LinearLayout, KoinComponent {
```
pantelisss, line 11: Are we sure that the header should be so big? I thought the title should be inside the scrolling area

pantelisss (changes requested):
Check my comment about the header. IMO is huge.

app/src/main/java/com/weatherxm/ui/components/HeaderView.kt
PavlosTze: Not sure really, I took it from the designs, and I thought due to the fact that it's a header, that it won't be scrollable. Let's discuss it internally with the product team.

pantelisss (approved):

JohnFrontzos (approved):

PR #57, opened 2024-03-06
FE-666 FE-664 FE-694 Feature/reward boosts

PavlosTze:
## **Why?**
Implement new reward boosts screen.

### **How?**
1. Created `RewardBoostActivity`, `RewardBoostViewModel` and `RewardBoostDetailAdapter` along with its respective XML files.
2. Created `UIBoost` which is the data class who we use to populate the boost screen's UI.
3. Function `getBoostReward` in `RewardsUseCase` does all the magic of creating that `UIBoost` based on the available info we have.

### **Testing**
Run on development environment and test the reward boost details against multiple stations. Station `Immense Nylon Altocumulus` is a station who has some dummy (not 0) data.


pantelisss (changes requested):
Looks good! Only one thing, the back button in the rewards boost screen doesn't work.
The hardware back works but the arrow button no

PavlosTze:
Good catch Panteli, fixed!

app/src/main/java/com/weatherxm/usecases/RewardsUseCase.kt
```diff
@@ -19,9 +31,17 @@ interface RewardsUseCase {
...
     ): Either<Failure, RewardDetails>
+
+    suspend fun getBoostReward(
+        deviceId: String,
+        boostReward: BoostReward
+    ): Either<Failure, UIBoost>
 }
 
-class RewardsUseCaseImpl(private val repository: RewardsRepository) : RewardsUseCase {
+class RewardsUseCaseImpl(
```
JohnFrontzos, line 41: Am not sure if the UI's Resources class should be used inside a UseCase class. It's a transformation to create a "State UI" model that I would consider moving to ViewModel. Or you can create a model without Resource IDs and in ViewModel class you can add all the contexted-based staff.

PavlosTze: I remember when I had some architecture-related discussions back then, that one functionality of a UseCase would be to prepare the UI Model for the upper layers (ViewModel) and the ViewModel to prepare / have available anything else it might need to the activity.

That's why I defined this model here, what do you think of that? Shall we create the UI Model in the ViewModel?

Btw we have used this pattern in other UseCases as well afaik, but of course if we decide to change this pattern we can optimize it in others as well (in the future).

PR #59, opened 2024-03-08
FE-577 When current weather data is null, show the respective component

PavlosTze:
## **Why?**
"No Data" view in device details when weather data is empty or null.

### **How?**
1. Added the respective component in .XML
2. Make the respective checks in `CurrentFragment` and `WeatherCardView` in order to handle visibility changes in each case.

### **Testing**
Try local mock flavor and delete the current weather object in `get_user_device.json` or find such an empty station in our network and test against it.


pantelisss (approved):
👌

JohnFrontzos (approved):

PR #72, opened 2024-03-12
FE-706 Feature/cells deep link

PavlosTze:
## **Why?**
Handle deep linking of `https://explorer.weatherxm.com/cells/*` URLs.

### **How?**
1. Added the respective `data` block in the `UrlRouterActivity` in the `AndroidManifest`.
2. Created `getCellInfo` function in UseCase which gets all cells and returns the one that matches the predefined `index` or a `CellNotFound` failure.
3. Implemented the "search and return a cell" functionality in `UrlRouterViewModel` in order to properly either fetch the cell and pass it to `CellInfoActivity` and open that activity, either return a failure and handle it accordingly.
4. Added a listener on back pressed on `CellInfoActivity` to open the explorer when this activity gets opened through a deep link
5. The "navigate to location" functionality on Explorer wasn't working, so I made the respective fixes on `ExplorerViewModel` and optimized the code there.

### **Testing**
**Use Prod flavor:**
1. Either use the "Share Cell" feature in the app or open cell details through the explorer on any cell in your mobile browser
7. Copy the link and open it in a browser or send it via any Instant Messaging to yourself and try opening it directly (e.g. through Messenger/Telegram/Slack etc.)
8. A pop-up should show up as a prompt to open the page in-app. If not, check if there is an "Open in app" option by clicking on the 3 dots at the top right corner of the browser/page
9. If still nothing is there, ensure that the app has permissions to open `explorer.weatherxm.com` URLs (you can find the option through app settings in your device). Enable the permissions and go back to step 2.

Ensure that the above flow is working properly.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -136,6 +136,11 @@
                     android:host="${explorer_host}"
                     android:pathPattern="/stations/.*"
                     android:scheme="https" />
+
+                <data
+                    android:host="${explorer_host}"
+                    android:pathPattern="/cells/.*"
+                    android:scheme="https" />
             </intent-filter>
         </activity>
         <activity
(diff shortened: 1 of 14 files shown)


pantelisss (approved):
👌

PR #89, opened 2024-03-27
FE-740 FE-737 FE-731 FE-728 Feature/update reward score colors icons

PavlosTze:
## **Why?**
Update reward score colors to 3-color scheme. Update base reward and boost icons.

### **How?**
Created a new function `getRewardIcon` which returns the correct icon based on the reward score (checkmark, warning, tint) and made the respective changes in `getRewardScoreColor` to return the correct colors.

### **Testing**
Just ensure everything is according to the spec color/icon wise.


app/src/main/java/com/weatherxm/ui/components/DailyRewardsCardView.kt
```diff
@@ -74,13 +81,23 @@ open class DailyRewardsCardView : LinearLayout, KoinComponent {
...
+            binding.boostsIcon.setColor(R.color.midGrey)
+            binding.boosts.setVisible(false)
+            binding.noActiveBoosts.setVisible(true)
+        } else {
+            binding.boostsIcon.setImageDrawable(
+                AppCompatResources.getDrawable(context, R.drawable.ic_checkmark_hex_filled)
+            )
+            binding.boostsIcon.setColor(R.color.network_stats_chart_primary)
+        }
         data.totalBoostReward?.let {
```
pantelisss, line 97: I think there is something wrong here in the recycler view. Take a look at the screenshot. If you scroll until the first "No active boosts" transaction and then scroll to the top, there are cards with checkmark icon along with "No Active Boosts"
Screenshot_20240327_151640

PavlosTze:
Good catch @pantelisss , fixed. Please check again. I also fixed FE-728, it was due to the same cause.

pantelisss (approved):
🚀

PR #90, opened 2024-03-27
FE-750 Use CLAIM_APP_URL from environment secret

PavlosTze:
## **Why?**
Add a dev URL for the claim app to dev builds.

### **How?**
Changed the GitHub workflows to use the environment secret for `CLAIM_APP_URL` and set those secrets in this repo.

### **Testing**
Not testable until merged.

### DIFF
--- a/.github/workflows/on-qa-firebase-distribution.yml
+++ b/.github/workflows/on-qa-firebase-distribution.yml
@@ -16,8 +16,6 @@ on:
                     - "debug"
                 default: 'release'
 env:
-    # Claim DApp URL
-    ORG_GRADLE_PROJECT_CLAIM_APP_URL: ${{ secrets.CLAIM_APP_URL }}
     # Mapbox SDK secrets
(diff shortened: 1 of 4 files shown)


.github/workflows/on-qa-firebase-distribution.yml
```diff
@@ -38,6 +36,8 @@ jobs:
         env:
             # API URL for the environment selected
             ORG_GRADLE_PROJECT_API_URL: ${{ secrets.API_URL }}
+            # Claim DApp URL
+            ORG_GRADLE_PROJECT_CLAIM_APP_URL: ${{ secrets.CLAIM_APP_URL }}
```
pantelisss, line 40: Is it worth declaring this claim app URL as input with 2 options (`claim` and `claim-app-dev`)?

PavlosTze: Hmm I don't think we need it, as we are playing with environment secrets. It's definitely easier having two different options (with dev suffix) but we had structured already the building process to use only one variable (as in `API_URL` also) exactly for that reason, to not have multiple variables for the same thing.

pantelisss (approved):

PR #92, opened 2024-03-29
FE-754 Limit Reward Timeline data

PavlosTze:
## **Why?**
Limit timeline data to 3 months in the past. Some UI fixes and added an end of data card.

### **How?**
- Create a new object TimelineReward for a better approach based on Pantelis comment below. 
- When we get to the end of the data, we add a new `TimelineReward` with `type = END_OF_DATA` at the end of the list which the adapter handles to show the respective "end of data card" at the end of the list.
- Removed the functionality of `timezone`, `fromDate`, `toDate` in the API call in the ViewModel and UseCase as these are not needed anymore and it's part of the `Repository` to create the proper `fromDate` for the DataSource etc.

### **Testing**
Check any station's reward timeline and ensure everything is OK and you can see the "End of data" card at the 3-months mark.

### DIFF
--- a/app/src/main/java/com/weatherxm/ui/rewardslist/RewardsListViewModel.kt
+++ b/app/src/main/java/com/weatherxm/ui/rewardslist/RewardsListViewModel.kt
@@ -6,7 +6,6 @@ import androidx.lifecycle.ViewModel
 import androidx.lifecycle.viewModelScope
 import com.weatherxm.data.Failure
 import com.weatherxm.data.Resource
-import com.weatherxm.data.Reward
 import com.weatherxm.ui.common.RewardTimelineType
 import com.weatherxm.ui.common.TimelineReward
 import com.weatherxm.usecases.RewardsUseCase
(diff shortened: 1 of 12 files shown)


app/src/main/java/com/weatherxm/ui/rewardslist/RewardsListViewModel.kt
```diff
@@ -66,66 +56,21 @@ class RewardsListViewModel(
...
+                    onNewRewardsPage.postValue(Resource.success(currentShownRewards))
                 }.onLeft {
                     analytics.trackEventFailure(it.code)
                 }
+
                 blockNewPageRequest = false
             }
+        } else if (!hasNextPage) {
+            // Add an empty reward to indicate the end of data and send all rewards to the activity
+            currentShownRewards.add(Reward.empty())
```
pantelisss, line 72: Just a suggestion because I don't like these implicit solutions.
Maybe you can declare a new object with two properties 
`type`: with values `data` and `endOfList`
`data`: `Reward` which will be null if the `type` is `endOfList`

pantelisss (commented):
Just check my suggestion. If you like it, we can follow this approach.
Otherwise, we are good to go

app/src/main/java/com/weatherxm/ui/rewardslist/RewardsListViewModel.kt
PavlosTze: Hmm I think that's a valid suggestion. I'm waiting for feedback from @JohnFrontzos (please see Pantelis' suggestion as well, I think his suggestion aligns with yours) before proceeding to implementing this.

JohnFrontzos: Yes, I agree with @pantelisss.

pantelisss (approved):

PR #98, opened 2024-04-01
FE-753 Feature/low battery station details

PavlosTze:
## **Why?**
Integrate low battery warning in station details.

### **How?**
- Introduce the new `Chip` in the layout file, along with `ChipGroup` and `HorizontalScrollView`.
- Created a new function in `DeviceDetailsActivity` named `setAlerts` which handles the text/visibility of this "alert" `Chip`
- Removed the overflow cards below the weather card except the `offline` one where we will keep the same logic.
- When the device is offline make the status chip clickable, upon click the app navigates to the Alert screen.
- Add the respective code in `DeviceDetailsUseCaseImpl` to return the "Low Battery" alert (currently will return it **in all** owned devices as the testing code is there but marked with a `#FIXME` comment)

### **Testing**
Ensure that all 3 different type of alerts are shown correctly in the station details screen with any possible combination,

### DIFF
--- a/app/src/main/java/com/weatherxm/ui/devicedetails/DeviceDetailsActivity.kt
+++ b/app/src/main/java/com/weatherxm/ui/devicedetails/DeviceDetailsActivity.kt
@@ -15,6 +15,7 @@ import androidx.lifecycle.repeatOnLifecycle
 import androidx.viewpager2.adapter.FragmentStateAdapter
 import com.google.android.material.dialog.MaterialAlertDialogBuilder
 import com.google.android.material.tabs.TabLayoutMediator
+import com.google.firebase.analytics.FirebaseAnalytics
 import com.weatherxm.R
 import com.weatherxm.data.Resource
 import com.weatherxm.data.Status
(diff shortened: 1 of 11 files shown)


app/src/main/java/com/weatherxm/ui/devicedetails/DeviceDetailsActivity.kt
```diff
@@ -250,12 +254,41 @@ class DeviceDetailsActivity : BaseActivity() {
...
+        } else if (alertsWithoutOffline.size == 1) {
+            if (alertsWithoutOffline[0].alert == DeviceAlertType.NEEDS_UPDATE) {
+                binding.alertChip.text = getString(R.string.update_required)
+            } else if (alertsWithoutOffline[0].alert == DeviceAlertType.LOW_BATTERY) {
+                binding.alertChip.text = getString(R.string.low_battery)
+            }
+            binding.alertChip.setVisible(true)
+        } else {
+            binding.alertChip.setVisible(false)
+        }
```
pantelisss, line 291: Minor, maybe you can use something like this under the `if`s to make things more clear
```
binding.alertChip.setVisible(!alertsWithoutOffline.isEmpty())
```

pantelisss (approved):
Looks good! Just check my suggestion

app/src/main/java/com/weatherxm/ui/devicedetails/DeviceDetailsActivity.kt
PavlosTze: Good catch Panteli, you are right! I just fixed it.

PR #99, opened 2024-04-02
FE-757 FE-749 Feature/release builds flow

PavlosTze:
## **Why?**
Fixes in the release builds flow (especially on dev and production releases)

### **How?**
- Replaced `FIREBASE_PUBLIC_TEST_GROUP` and `FIREBASE_INTERNAL_TEST_GROUP` with the generic `FIREBASE_TEST_GROUP` which we define it differently via the secrets `FIREBASE_TECH_TEAM_TEST_GROUP` and `FIREBASE_QA_TEST_GROUP`.
- Do **NOT** send anymore production builds when pushing on `main`
- Introduce a new secret named `ENCODED_GPLAY_SERVICE_ACCOUNT` which is the service account required for Google Play deployment, we decode it by doing `echo "${{ secrets.ENCODED_GPLAY_SERVICE_ACCOUNT }}" | base64 -d > ./gplay-ci-service-account.json` and then we build the `AAB` file and deploy it using [upload-google-play](https://github.com/marketplace/actions/upload-android-release-to-play-store).

### **Testing**
I'm not sure it's testable right now, I think it needs to get merged first so have a look at the changes and ensure everything looks good.


pantelisss (approved):

JohnFrontzos (approved):
gg wp 🎉

PR #112, opened 2024-04-15
FE-785 Feature/redesign forecast tab

PavlosTze:
## **Why?**
Redesign of Forecast Tab.

### **How?**
- Created a specific `ForecastUseCase` to handle everything Forecast-related and remove that info from `DeviceDetailsUseCase`
- Created the respective `UIModels` in order to populate the screen.
- Replaced precipitation icon
- Replaced `HourlyAdapter` with `HourlyForecastAdapter` which is the adapter handling the top hourly tiles
- Replaced `getFormattedRelativeDay` with `getRelativeDayOrFull` and `getRelativeDayAndMonthDay` in order to reuse them respectively.
- Changes in the respective forecast-related XML files and the `DailyForecastAdapter` and `ForecastFragment`

### **Testing**
Ensure everything looks OK in the prod or dev environment. The only thing missing is that the start of the hourly forecast does **not** start with the current hour as the API hasn't implemented it yet.


app/src/main/java/com/weatherxm/usecases/ForecastUseCaseImpl.kt
```diff
@@ -0,0 +1,63 @@
...
+
+    @Suppress("MagicNumber")
+    override suspend fun getForecast(
+        device: UIDevice,
+        forceRefresh: Boolean
+    ): Either<Failure, UIForecast> {
+        if (device.timezone.isNullOrEmpty()) {
+            return Either.Left(ApiError.UserError.InvalidTimezone(INVALID_TIMEZONE))
+        }
+        val nowDeviceTz = ZonedDateTime.now(ZoneId.of(device.timezone))
```
pantelisss, line 28: IMO is better to get the timezone from the forecast response instead of injecting the device object

app/src/main/java/com/weatherxm/usecases/ForecastUseCaseImpl.kt
```diff
@@ -0,0 +1,63 @@
...
+            nowDeviceTz,
+            dateEndInDeviceTz,
+            forceRefresh
+        ).map { result ->
+            val nextHourlyWeatherForecast = mutableListOf<HourlyWeather>()
+            val forecastDays = result.map { weatherData ->
+                weatherData.hourly?.filter {
+                    val isCurrentHour = it.timestamp.dayOfYear == nowDeviceTz.dayOfYear
+                        && it.timestamp.hour == nowDeviceTz.hour
+                    isCurrentHour || it.timestamp.isAfter(nowDeviceTz)
```
pantelisss, line 41: Can we do something like the following?
```
it.timestamp >= nowDeviceTz && it.timestamp < nowDeviceTz.plusHours(24)
```
And omit picking the first 24 items. With this, we handle the case of gaps (if there are for any reason) and the case of non-hourly samples

pantelisss (changes requested):
Check my comments

app/src/main/java/com/weatherxm/usecases/ForecastUseCaseImpl.kt
PavlosTze: This is true but we need this `nowDeviceTz` in order to create the `dateEnd` for the API call below:
```
        val nowDeviceTz = ZonedDateTime.now(ZoneId.of(device.timezone))
        val dateEndInDeviceTz = nowDeviceTz.plusDays(7)
        return weatherForecastRepository.getDeviceForecast(
            device.id,
            nowDeviceTz,
            dateEndInDeviceTz,
            forceRefresh
        )
```
So we need the timezone before we have the actual response that's why we inject the device object.

app/src/main/java/com/weatherxm/usecases/ForecastUseCaseImpl.kt
PavlosTze: Valid point, and it's much cleaner now. Thanks!

pantelisss (approved):
👌

PR #117, opened 2024-04-17
FE-786 Forecast details

PavlosTze:
## **Why?**
The new forecast details screen and its respective components + optimizations due to that new screen which led to pieces of the code that could be optimized to reused around the app.

### **How?**
- Created a `ChartsUseCase` which is responsible for preparing the chart-related UI Model to any ViewModel that needs it (and removed the respective functionality from `HistoryUseCase`.
- Created the forecast details layout file along with its activity and view model. We pass to the activity the forecast data, the selected day (which is nullable) and the selected hour (which is also nullable) in order to know which day/hour to scroll and show in the screen.
- Created the `DailyTileForecastAdapter` and its view component which is an adapter capturing the functionality of the daily tiles in the forecast details screen.
- Created a new `WeatherMeasurementCard` component to populate the "small" weather cards in the forecast details and reuse them whenever we can.
- Created a new extension function `Lottie.setWeatherAnimation` in order to reuse it whenever we can and not repeat the same code throughout the app.
- Removed the `HistoryUIModels` and moved these models in the `common/UIModels` and renamed it to `Charts` as we reuse them in History & Forecast.
- Created extension function `String.boldText` to reuse it throughout the app and make the respective text bold.
- Created extension function `TextView.setDisplayTimezone` to reuse it as we have the same code in History & Forecast and in the future possibly in other screens as well.
- Added some functions in `LineChartView` in order to be able to programmatically customize some components.
- Created a new `ChartsView` component which is a view containing all the charts and encapsulating their code, and made that a bit more agnostic in order to reuse this code in Forecast & History. Removed the respective explicit code from `HistoryFragment` and `HistoryViewModel`.
- Fixed `getFormattedWindDirection` to accept null values (for code simplicity).
- Moved `getWindDirectionDrawable` from `Resources` to `Weather` and use `AppCompatResources`.
- Created a `getUVClassification` function which returns the result of the UV mapping.
- Function `setData` in `WeatherDataView` now accepts a nullable drawable to update its icon - currently used only in wind in order to show direction.
- Fixes in `Charts` util functions in order to make them more reusable when needed.
- Created `getRelativeDayAndShort` function which returns dates in the format `Today, Wed, Apr 2` or `Wed, Apr 2`

### **Testing**
In either prod or dev flavor, ensure the forecast details screen is OK in both themes. Also check that the Historical charts work ok as well.


pantelisss (changes requested):
Good job!
2 comments 
- Hourly forecast scroller should go back to 7 o'clock on every change from top picker
- If it's easy, scroll to temperature graph on main card tap

pantelisss (approved):
👏👏👏

PR #130, opened 2024-04-30
FE-836 Feature/forecast improvements

PavlosTze:
## **Why?**
Fixed the issues as mentioned in the respective Linear task.

### **How?**
Most fixes where xml related, the only "new" thing is the introduction of a new function `screenLocation` which returns the `(x, y)` point of a `View`. We use this function in order to smooth scroll towards the chart we want.

### **Testing**
Make sure all issues have been fixed and the `[Nice to have]` has been implemented.


app/src/main/java/com/weatherxm/ui/deviceforecast/ForecastDetailsActivity.kt
```diff
@@ -189,7 +190,8 @@ class ForecastDetailsActivity : BaseActivity() {
     }
 
     private fun scrollToChart(chart: LineChartView) {
-        chart.parent.requestChildFocus(chart, chart)
+        val (x, y) = chart.screenLocation()
+        binding.mainContainer.smoothScrollTo(x, y - binding.appBar.height - binding.root.paddingTop)
```
pantelisss, line 194: This works properly only when the scroll view is scrolled on top. If you scroll a little  and tap on one of the items the final scroll position is wrong
Seems that something like this works.
```
val (x, y) = chart.screenLocation()
val currentY = binding.mainContainer.scrollY
binding.mainContainer.smoothScrollTo(x, y - binding.appBar.height - binding.root.paddingTop + currentY)
```

pantelisss (changes requested):
Everything looks nice, except for the issue with the scroll position. 
Check my comment

app/src/main/java/com/weatherxm/ui/deviceforecast/ForecastDetailsActivity.kt
PavlosTze: Good catch Panteli, fixed!

pantelisss (approved):
🚀

PR #131, opened 2024-04-30
FE-839 Include offline alert in count

PavlosTze:
## **Why?**
Include offline alert in count

### **How?**
Use all alerts in the counting chip and not the filtered ones.

### **Testing**
Testing code has been included to show all 3 different issues in Device Details so just ensure that "3 issues" is visible.


pantelisss (changes requested):
The issues chip should follow the severity logic. For example it the following screenshot the "3 issues" chip should be red
Screenshot 2024-05-01 at 10 41 00

pantelisss (approved):

PR #132, opened 2024-05-01
FE-817 Integrate Mixpanel Analytics

PavlosTze:
## **Why?**
Integrate Mixpanel Analytics.

### **How?**
- Add `MIXPANEL_TOKEN` through the `production.env` file as a `BuildConfig` value and updated the `production.env.template` file.
- On every different method in `Analytics` made the respective adjustments so each reporting function now reports at both Firebase and Mixpanel.
- Created for this purpose 2 functions, the first one is `paramsToJSON` which is a function which takes as argument the list of key,value pairs and returns a JSONObject and the second one is `logEvent` which logs the different events in Firebase and Mixpanel.

### **Testing**
1. Add `MIXPANEL_TOKEN` in the `production.env` file as shown on the updated template
2. Deploy on your device a "releaseProd" flavour, navigate in the app and then check if the corresponding events are shown in the Mixpanel dasboard.


PavlosTze:
This is ready for testing 😃

JohnFrontzos (approved):

pantelisss (approved):
👏

PR #144, opened 2024-05-09
FE-824 Refactor Claiming - Connect to WiFi

PavlosTze:
## **Why?**
Connect the gateway to your WiFi part.

### **How?**
- Refactored the whole "WiFi" logic in order to have only one "ClaimWifiXXX" activity/fragments and not different for M5 and D1.
- Updated the content of "Connect to WiFi" step.
- Updated the steps in the WiFi claiming flow by replacing the chips with the new progress bar.

### **Testing**
Open both M5 and D1 claiming flows and ensure everything is OK.


PavlosTze:
Now ready for review @pantelisss / @JohnFrontzos

pantelisss (changes requested):
Check my comments. Also moving back from the claim device screen goes to the Home Screen instead of the device selector.

pantelisss (approved):
👌

PR #150, opened 2024-05-14
FE-850 Catch a rare NotConnectedException when disconnecting bluetooth

PavlosTze:
### **How?**
Catch the `NotConnectedException` in order to avoid crashes.

### **Testing**
Edge case, I'm not sure if there is a point in testing it or how to.


pantelisss (approved):
Code-wise makes sense 🙂

PR #151, opened 2024-05-14
FE-827 Feature/refactor claim set location

PavlosTze:
## **Why?**
Claim Wizard step 4: Confirm station Location.

### **How?**
Created a different M5/D1 container in the same layout as the current "set location" screen and toggle visibility & handling in the Fragment depending on which `DeviceType` we have.

This is a **temp & simple workaround** until the design team updates the Helium screens as well, which is already WIP.

### **Testing**
Ensure that the M5/D1 step matches the design and the Helium's respective step is OK as well. Also try some claiming tests in M5 and D1 and ensure everything is OK.


pantelisss (approved):
🫡

PR #174, opened 2024-05-31
FE-884 Update Helium's claiming flow to the new designs

PavlosTze:
## **Why?**
Refactor the helium logic and migrate it with the newly implemented M5/D1 logic.

### **How?**
- UI fixes here and there in all steps to ensure that the texts/margins/fonts/colors etc are the same as the designs

### **Testing**
Ensure that everything is OK in the Helium's claiming flow.

### DIFF
--- a/app/src/main/res/layout/fragment_claim_helium_reset.xml
+++ b/app/src/main/res/layout/fragment_claim_helium_reset.xml
@@ -20,14 +20,13 @@
 
             <com.google.android.material.textview.MaterialTextView
                 android:id="@+id/resetTitle"
-                android:layout_width="wrap_content"
+                android:layout_width="match_parent"
                 android:layout_height="wrap_content"
-                android:text="@string/reset_station"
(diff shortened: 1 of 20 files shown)


app/src/main/res/layout/fragment_claim_helium_reset.xml
```diff
@@ -20,14 +20,13 @@
 
             <com.google.android.material.textview.MaterialTextView
                 android:id="@+id/resetTitle"
-                android:layout_width="wrap_content"
+                android:layout_width="match_parent"
                 android:layout_height="wrap_content"
-                android:text="@string/reset_station"
+                android:text="@string/reboot_your_device"
```
pantelisss, line 25: This was by mistake
It should be "Reset your device", check the updated designs

pantelisss (changes requested):
Codewise looks good! Just fix my the title and we're good to go

pantelisss (approved):

PR #189, opened 2024-06-06
FE-900 Integrate some claiming flow analytics

PavlosTze:
## **Why?**
Capture some important KPIs (how many users started the claiming process and the result of it - success or failure).

### **How?**
- Track the ScreenView in the `onResume` at `RewardsClaimActivity`
- Based on the result of the claiming in `ProfileFragment` track a `VIEW_CONTENT` event containing the `STATE` (success or failure, for now) parameter.

### **Testing**
Run a **release** flavor (not debug), enable analytics and ensure that events are logged properly.


pantelisss (approved):
🤌

PR #191, opened 2024-06-07
FE-911 Fix validation of serial number & claiming key inputs

PavlosTze:
## **Why?**
Fix validation of serial number & claiming key inputs as it allows to proceed with an invalid serial number by using the "Enter" key on the keyboard.

### **How?**
Refactor how we validate the input fields as before we didn't "terminate" the execution of the `fun` when an error was found.

### **Testing**
Ensure that it's not possible to proceed either with the button or with the enter in the keyboard when the serial number and/or the claiming key are invalid. Also check M5's manual input.


pantelisss (approved):
🫡

PR #198, opened 2024-06-11
FE-924 Integrate Claiming Pulse Step 2 - Prepare gateway + Manual SN

PavlosTze:
## **Why?**
Prepare Gateway + Scanning Barcode + Manual Details

### **How?**
- Created `ClaimPulsePrepareGatewayFragment` and `ClaimPulseManualDetailsFragment` with its respective layouts and added them in the Pager
- Created helper function `removePulsePrefix` which removes the prefix "P" **if** it exists, otherwise returns the string as-is

### **Testing**
Ensure that everything looks OK and the validation of the input works correctly.


pantelisss (approved):
👌

PR #202, opened 2024-06-11
FE-912 Add Mapbox style as env param

PavlosTze:
## **Why?**
Use standard MapBox style as default, and add an env param for WeatherXM-themed map style. Also, pass the env param in GitHub CI.

### **How?**
- Inject our style via `production.env` param `MAPBOX_STYLE`
- On `ExplorerMapFragment` @ `getMapStyle` check if there is a custom style injected available otherwise use the default ones found in `BaseMapFragment`

### **Testing**
Ensure that the explorer looks the same as before **after** injecting the `MAPBOX_STYLE` param via `production.env` (follow the template as seen in this PR to add it).


pantelisss (approved):
🤝

PR #203, opened 2024-06-11
FE-936 Add support for AS923_1B and map it to Malaysia

PavlosTze:
### **How?**
- Added the respective entry in `enum class Frequency` and in the `frequencyToHeliumBleBandValue` mapping function.
- Map Malaysia to `AS923_1B`.

### **Testing**
The `AS923_1B` should be available in the frequencies list in the Change Frequency screen.


pantelisss (approved):
🇲🇾

PR #213, opened 2024-06-17
FE-948 Use Plurals and get them as quantity strings where applicable

PavlosTze:
## **Why?**
Use Plurals and get them as quantity strings where applicable.

### **How?**
- Created a `plurals.xml` values file where we setup our new quantity strings
- Removed the now deprecated strings from `strings.xml`
- Use `getQuantityString` function as described [here](https://stackoverflow.com/questions/41950952/how-to-use-android-quantity-strings-plurals](https://stackoverflow.com/questions/41950952/how-to-use-android-quantity-strings-plurals) in order to get the correct string which eventually reduces the number of `if...else` checks and simplifies the code.

### **Testing**
1. Ensure that the following screens work correctly for 0, 1 and more than 1 alerts/issues. Try either against the API or by playing with local mock.
    - Rewards Timeline - `get_device_rewards_timeline.json`
    - Reward Details -  `get_device_reward_details.json`

2. Ensure that the "Active Stations" in the cell details work correctly for 0, 1 and more than 1 stations.

### DIFF
--- a/app/src/local/assets/mock_files/get_device_rewards_timeline.json
+++ b/app/src/local/assets/mock_files/get_device_rewards_timeline.json
@@ -15,7 +15,14 @@
             "base_reward_score": 50,
             "annotation_summary": [
                 {
-                    "severity_level": "WARNING",
+                    "severity_level": "INFO",
+                    "group": "SENSOR_PROBLEMS",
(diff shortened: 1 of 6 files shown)


pantelisss (approved):
👌

PR #214, opened 2024-06-18
FE-955 Feature/use set visible function everywhere

PavlosTze:
## **Why?**
Use everywhere `setVisible` and `setInvisible` functions.

### **How?**
Replaced:
1. `View.VISIBLE` ---> `setVisible(true)`
2. `View.GONE` ----> `setVisible(false)`
3. `View.INVISIBLE` ----> `setInvisible()` (this is a new extension function created for this purpose)

### **Testing**
Ensure that the changes in the code follow the _"How"_ as described above.

### DIFF
--- a/app/src/main/java/com/weatherxm/ui/cellinfo/CellDeviceListAdapter.kt
+++ b/app/src/main/java/com/weatherxm/ui/cellinfo/CellDeviceListAdapter.kt
@@ -1,7 +1,6 @@
 package com.weatherxm.ui.cellinfo
 
 import android.view.LayoutInflater
-import android.view.View
 import android.view.ViewGroup
 import androidx.recyclerview.widget.DiffUtil
 import androidx.recyclerview.widget.ListAdapter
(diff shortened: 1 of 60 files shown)


pantelisss (changes requested):
IMO the naming is a little confusing.
As far I can  understand the `setVisible` shows the element when is `true` and **removes** it from the UI when is `false`. The `setInvisible` just hides the element which was the thing I except for the `false` in `setVisible`.
Maybe something like `keepOnView` instead of `setVisible` may be more explanatory
In any case, I'm not strong. Up to you

PavlosTze:
The `setInvisible` @pantelisss hides the view but not its space in the View, it's like "hiding the top of the UI but not the underlying container", as it uses `View.INVISIBLE` functionality instead of `View.VISIBLE` and `View.GONE` where appearing/disappearing of everything takes place.

So, `keepOnView` makes more sense for `setInvisible` than `setVisible`.

Do you have any better ideas for renaming `setInvisible`? I named it that way due to the fact that `Invisible` is a term in Android view system. If you think that sounds OK now after the explanation let's proceed this way.

pantelisss (approved):
👌

PR #217, opened 2024-06-20
FE-962 Feature/fix claiming

PavlosTze:
## **Why?**
Fixed 2 issues reported by the product team, described in the respective [Linear issue](https://linear.app/weatherxm/issue/FE-962/improvements-for-android-via-product-testing-phase).

### **How?**
- Increased the default zoom level from 1 to 2 on Mapbox as it seems due to a library updated they treat differently now the "1" we had before showing the Globe.
- Used the `anchorView` in the snackbar in order to show it above the buttons and added ad `dismiss` button.

### **Testing**
Ensure that both issues are fixed.


pantelisss (commented):
Codewise looks good. I couldn't test on a real device.
My only concern is if we should anchor the snackbar above the bottom buttons in every step

PavlosTze:
Hmm, the only other point where a snackbar could be applicable in this flow is in the location step, right?

Although it shouldn't happen only on an edge/buggy case. 

But I can move it above the buttons I think if needed.

PavlosTze:
I added the snackbar above the button in the location step as well.

pantelisss (approved):

PR #225, opened 2024-07-01
FE-984 Validation of friendly name

PavlosTze:
## **Why?**
Correct validation of friendly name.

### **How?**
Match when the following are met:
- any char
- min length = 1
- max length = 24
- non empty text (i.e. `return false` when only white space characters are used), new Regex `^\\S.{1,24}$`

#### Other:
Added the below code in order to remove the error shown when the user is typing and disable the "Save" button when the input is empty.
```
binding.newName.onTextChanged {
    binding.newNameContainer.error = null
    binding.save.isEnabled = it.isNotEmpty()
}
```

### **Testing**
Ensure that strings longer than >24 chars are not validating correctly and that only non-empty strings are matched.


pantelisss (approved):
👌

PR #226, opened 2024-07-02
Fix Regex of Friendly Name allowing 1 char

PavlosTze:
## **Why?**
Fix Regex of Friendly Name allowing 1 char

### **How?**
Replaced the regex with the correct one.

### **Testing**
Ensure that name is between 1 and 24 chars (not less, not more).


pantelisss (approved):
👌

PR #228, opened 2024-07-04
FE-899 FE-991 MVP of Unit Tests & Coverage with JaCoCo

PavlosTze:
## **Why?**
We need some first basic unit tests for each "different domain" (e.g UseCases, util functions, ViewModels, Repositories etc) which will run on every PR and a basic coverage reporting using JaCoCo.

## **How?**
- Added [Kotest](https://kotest.io/) library for Kotlin Unit Tests, [Mockk](https://mockk.io/) library for mocking and `androidx.arch.core.testing` which is a library handling `LiveData` in Unit Tests. We use **[Behavior Spec](https://kotest.io/docs/framework/testing-styles.html#behavior-spec)** in the way we write our unit tests which helps being self-descriptive.
- On PR CI/CD added another action to run unit tests with the `SKIP_TAGS_LOGGING` project parameter which skips logging the tags in order for the console to be as clear as possible. Added support for this parameter in `build.gradle`. Also added a "Test Report" step which can be found in each PR action.
- In `Validator` replaced `Patterns.EMAIL_ADDRESS.matcher(username).matches()` with `PatternsCompat` which provides better support and can be used in unit tests.
- Created various Unit Tests for different domains in our code which use the `shouldBe` assertion and `verify` or `coVerify` to ensure that a X `fun` has been invoked `exactly = Y` times.
- Created a Coverage GitHub Action which uses [JaCoCo](https://github.com/jacoco/jacoco) and the [jacoco-report action](https://github.com/Madrapps/jacoco-report) that prints in each PR coverage-related information.

## **Testing**
Just run the Unit Tests locally by right clicking on "test [unitTest" and then clicking "Run Tests":
εικόνα

Also if the unit tests that will run in this PR through CI/CD are successful.


PavlosTze:
@pantelisss This is ready for testing/review, low priority though as we need to wait for John as well to review it. Feel free to ping me for anything.

app/src/test/java/com/weatherxm/data/repository/FollowRepositoryTest.kt
```diff
@@ -0,0 +1,74 @@
...
+                }
+            }
+            When("it's valid") {
+                then("return a success") {
+                    repo.followStation("testId") shouldBe Either.Right(Unit)
+                    coVerify(exactly = 1) { networkSource.followStation("testId") }
+                }
+                and("Save it in the cache") {
+                    coVerify(exactly = 1) { cacheSource.followStation("testId") }
+                }
```
pantelisss, line 47: Maybe you should check if the device is saved in cache too

app/src/test/java/com/weatherxm/data/repository/FollowRepositoryTest.kt
```diff
@@ -0,0 +1,74 @@
...
+            }
+            When("it's valid") {
+                then("return a success") {
+                    repo.unfollowStation("testId") shouldBe Either.Right(Unit)
+                    coVerify(exactly = 1) { networkSource.unfollowStation("testId") }
+                }
+                and("Remove it from the cache") {
+                    coVerify(exactly = 1) { cacheSource.unfollowStation("testId") }
+                }
```
pantelisss, line 70: Same here, if the device is removed from the cache

pantelisss (changes requested):
Great job, just check my suggestions

app/src/test/java/com/weatherxm/data/FrequencyHelperTest.kt
```diff
@@ -0,0 +1,136 @@
...
+            When("AS923_3") {
+                then("Band Value = 12") {
+                    frequencyToHeliumBleBandValue(Frequency.AS923_3) shouldBe 12
+                }
+            }
+            When("AS923_4") {
+                then("Band Value = 13") {
+                    frequencyToHeliumBleBandValue(Frequency.AS923_4) shouldBe 13
+                }
+            }
```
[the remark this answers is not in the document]
PavlosTze: Nice suggestion, yes I'm implementing this!

app/src/test/java/com/weatherxm/data/repository/FollowRepositoryTest.kt
PavlosTze: The plan @pantelisss is to create different unit tests for `CacheService` so that each unit test has its own purpose. So not needed here to check if the device is saved or removed from cache, what do you think?

app/src/test/java/com/weatherxm/data/repository/FollowRepositoryTest.kt
PavlosTze: Same as above.

app/src/test/java/com/weatherxm/data/repository/FollowRepositoryTest.kt
pantelisss: I see. Ok then!

pantelisss (approved):
🤘🚀🧨

JohnFrontzos (approved):

PR #232, opened 2024-07-11
FE-1004 Feature/notifications fcm api

PavlosTze:
# **Why?**
Communicate fcmToken-related info with the API based on: 

### Cases:
- **APP STARTUP:** Update the FCM token and send it to the server only if the user is logged in.
- **LOGIN:** Upload the FCM token after a successful login.
- **LOGOUT / DELETE ACCOUNT:** Delete the FCM token.
- **FCM REFRESH:** Update the FCM token and send it to the server only if the user is logged in.
- **JWT EXPIRED:** Delete the FCM token.

# **How?**
- Created `NotificationsDataSource` which is responsible for communicating the FCM-related actions with the API and the `NotificationsRepository` which is responsible for preparing the `installationId` and the `fcmToken`.
- Moved the functionality of getting the `installationId` to `AppConfigDataSource`
- Minor fix in `AuthRequestInterceptor` in order to avoid throwing an exception when `body == null` which is a common exception on every logout.
- On every login through `AuthUseCase`, on every manual logout through `PreferencesUseCase` and on every delete account through `DeleteAccountUseCase` we set and delete the FCM token to/from the server respectively.
- Created `RefreshFcmApiWorker` which is a Worker for [one-time work](https://developer.android.com/develop/background-work/background-tasks/persistent/getting-started/define-work#schedule_one-time_work) responsible for:
    - Updating the FCM token in the API asynchronously when the `onNewToken` is invoked in `MessagingService`.
    - Updating the FCM token in the API asynchronously on every startup if the user is logged in
- Created `ForceLogoutWorker` which is a Worker for [one-time work](https://developer.android.com/develop/background-work/background-tasks/persistent/getting-started/define-work#schedule_one-time_work) which took the functionality from `AuthTokenAuthenticator` (this authenticator is now responsible for starting this worker) and is responsible for forcing logout and clearing the respective cache, data and the FCM token.

# **Testing**
Make sure through the logger in Android Studio that the correct calls are made to the API endpoints even though the response should be 404 as these endpoints aren't implemented yet.


app/src/main/java/com/weatherxm/service/MessagingService.kt
```diff
@@ -51,7 +52,11 @@ class MessagingService : FirebaseMessagingService() {
      * FCM registration token is initially generated so this is where you would retrieve the token.
      */
     override fun onNewToken(token: String) {
-        Timber.d("Refreshed Firebase token: $token")
+        Timber.d("Refreshed FCM token: $token")
+        /**
+         * Init and invoke the work manager to update FCM token in the server
+         */
+        RefreshFcmApiWorker.initAndRefreshToken(applicationContext, token)
```
pantelisss, line 59: Maybe I'm wrong, but are we sure this object will not be deallocated from the memory before finishes thje job?
I'm saying this because no one holds a ref to this object

pantelisss (commented):
Overall, it looks nice. Just check my question, I'm not sure if is an issue

app/src/main/java/com/weatherxm/service/MessagingService.kt
PavlosTze: No, [Workers](https://developer.android.com/reference/androidx/work/Worker) are the proper way of doing work in the background in Android without having to rely on anything else, and we use `applicationContext` here as well.

Taken from the [documentation](https://developer.android.com/reference/androidx/work/Worker):

> A class that performs work synchronously on a background thread provided by [WorkManager](https://developer.android.com/reference/androidx/work/WorkManager).
> 
> If you need to do your work asynchronously or call asynchronous APIs, you should use [ListenableWorker](https://developer.android.com/reference/androidx/work/ListenableWorker).

Also another documentation here regarding [WorkManager](https://developer.android.com/reference/androidx/work/WorkManager):

> WorkManager is the recommended library for persistent work. Scheduled work is guaranteed to execute sometime after its [Constraints](https://developer.android.com/reference/androidx/work/Constraints) are met.
> 
> All work must be done in a [ListenableWorker](https://developer.android.com/reference/androidx/work/ListenableWorker) class. A simple implementation, [Worker](https://developer.android.com/reference/androidx/work/Worker), is recommended as the starting point for most developers. With the optional dependencies, you can also use CoroutineWorker or RxWorker. All background work is given a maximum of ten minutes to finish its execution. After this time has expired, the worker will be signalled to stop.

So it's safe to assume that this code will run no matter what :)

pantelisss: I see, WorkManager holds a ref on the task.

pantelisss: I think we are good to go

PavlosTze: Yup, ok. Let's wait for the backend to implement the API endpoint and review it again then?

PavlosTze:
@pantelisss This is ready to test it against the dev API, including the change we agreed onto today, regarding the `DELETE` request.

pantelisss (approved):
Seems nice! 👌

PR #239, opened 2024-07-15
FE-1013 Prevent a crash and print some logs in FirebaseCrashlytics

PavlosTze:
### **Why?**
Prevent a crash and print some logs in FirebaseCrashlytics so we can figure out the cases it happens in Forecast Details.

### **How?**
If `position == -1` (which produces a crash afterwards), catch it, return `0` as the fallback and print the error through `Timber.e` so we can try to reproduce it afterwards.

### **Testing**
Not testable as I haven't reproduced it yet. So I'm printing some logs in Firebase Crashlytics in case they help.

### DIFF
--- a/app/src/main/java/com/weatherxm/ui/deviceforecast/ForecastDetailsViewModel.kt
+++ b/app/src/main/java/com/weatherxm/ui/deviceforecast/ForecastDetailsViewModel.kt
@@ -94,7 +94,8 @@ class ForecastDetailsViewModel(
              * Temporary code so we can figure out why some crashes occur, in which cases/dates
              */
             val allDates = forecast.forecastDays.joinToString(" : ") { it.date.toString() }
-            Timber.e("Could not find selected day in forecast: $selectedISODate $allDates")
+            Timber.e("Could not find selected day " +
+                "($selectedISODate - $selectedLocalDate) in forecast: $allDates")
             0
         } else {
             position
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/weatherxm/ui/deviceforecast/ForecastDetailsViewModel.kt
```diff
@@ -80,13 +80,24 @@ class ForecastDetailsViewModel(
...
+        val position = forecast.forecastDays.indexOfFirst {
+            it.date == selectedLocalDate
+        }
+
+        return if (position == -1) {
+            /**
+             * Temporary code so we can figure out why some crashes occur, in which cases/dates
+             */
+            val allDates = forecast.forecastDays.joinToString(" : ") { it.date.toString() }
+            Timber.e("Could not find selected day in forecast: $selectedISODate $allDates")
```
pantelisss, line 97: Maybe we should include the `selectedLocalDate` as well

pantelisss (changes requested):
Looks good, just check my comment and we are good to go

app/src/main/java/com/weatherxm/ui/deviceforecast/ForecastDetailsViewModel.kt
PavlosTze: Valid point, done!

PR #241, opened 2024-07-16
FE-1015 Update Kable library and sort them alphabetically

PavlosTze:
### **Why?**
Update Kable library and sort them alphabetically

### **Testing**
1. Ensure that the project builds successfully
2. Connect to a device via BLE and perform an operation (e.g. Reboot)


pantelisss (approved):
Codewise seems ok and runs as expected. 
Unfortunately, I could not test it on a real device for the BLE check

PR #243, opened 2024-07-19
FE-1017 Feature/kable ble scanner

PavlosTze:
### **Why?**
Replace an obsolete library we were using with the Kable library for BLE scanning which we already use in connection & communication via BLE.

### **How?**
- Refactored (and simplified **a lot**) the `BluetoothScanner` class which is responsible for providing the `Flow` to the scanned devices
- Removed `registerOnScanning`, `startScanning` and `stopScanning` functions in Data Sources, Repositories and UseCases and used the `scan` method instead which exposes the above `Flow` and in the UseCase the scanned devices get transformed to `ScannedDevice` UI Model.
- Proper handling of scanning and getting the results in `ClaimHeliumPairViewModel`
- Proper handling of scanning, getting the results and connecting in `BluetoothHeliumViewModel`

### **Testing**
Ensure that BLE scanning & connection works OK in the following cases **for Helium**:
1. Claiming (no need to try for a success claim)
2. Reboot
3. Change Frequency

### **Additional context**
1. Removed the library `arrow-fx-coroutines = { module = "io.arrow-kt:arrow-fx-coroutines" }` which are not used in our project
2. Removed the custom implementation of the `reboot` command and used the standard `setATCommand` we use already in other cases, which works just fine for reboot now.


pantelisss (approved):
Codewise looks nice!

PR #246, opened 2024-07-22
FE-1029 Feature/user properties owned stations wallet

PavlosTze:
### **Why?**
Add in Analytics User Properties the `STATIONS_OWN` and `HAS_WALLET` params.

### **How?**
Save the respective entries in `SharedPreferences` and use them on `AnalyticsInitializer` in order to set them as user params.

### **Testing**
Ensure that these params are logged correctly **after** the first run (as the cache needs to be initialized first) by checking the logger in Android Studio.

### DIFF
--- a/app/src/main/java/com/weatherxm/analytics/AnalyticsService.kt
+++ b/app/src/main/java/com/weatherxm/analytics/AnalyticsService.kt
@@ -269,6 +269,8 @@ interface AnalyticsService {
         INCHES("in"),
         HPA("hpa"),
         INHG("inhg"),
+        STATIONS_OWN("STATIONS_OWN"),
+        HAS_WALLET("HAS_WALLET")
     }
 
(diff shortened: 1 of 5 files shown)


pantelisss (changes requested):
If I'm not missing something, I couldn't find somewhere the cleanup of these properties on logout

PavlosTze:
They are cleaned up in CacheService `clearAll` function:
```
    fun clearAll() {
        this.walletAddress = null
        this.user = null
        this.forecasts.clear()
        this.suggestions.clear()
        this.locations.clear()
        this.followedStationsIds = listOf()
        this.userStationsIds = listOf()

        okHttpCache.evictAll()
        encryptedPreferences.edit().clear().apply()
        clearUserPreferences()
    }
```

pantelisss (approved):

PR #267, opened 2024-08-09
FE-1087 Fix scanning barcodes in Pulse

PavlosTze:
### **Why?**
Use the ZXing library for scanning barcodes in Pulse. Disable auto zoom in Google Scanner.

### **How?**
- Integrated again the library we used before for scanning barcodes in Pulse
- Disabled auto-zoom in Google Scanner
- Use our automated flow for asking for permissions (in our case for `CAMERA` and handling denials etc.)

### **Testing**
Ensure that everything permissions-related is OK and scanning works properly in Pulse barcodes.


pantelisss (approved):
Codewise looks good. Could not tested on a real device

PR #275, opened 2024-08-13
FE-1033 FE-1035 Station Notifications

PavlosTze:
### **Why?**
Receive a push notification for a station, and navigate to the specific device. And Create a new category of notification called `Station Notifications` with ID `station`.

### **How?**
- Create `createPendingIntent` function in `MessagingService` which handles the remote message when the app is in the foreground/background and acts accordingly in each different case
- Create `showDeviceDetailsWithBackStack` function in `Navigator` which is the entry point to Device Details with the Home Screen as back stack.
- Support opening Device Details only with the `deviceID` instead of opening it before only with the `Device` object (as with notifications we have only the `deviceID`
- Support opening that new notification in `StartupActivity` which launches when a notification is clicked and the app is closed

### **Testing**
Test it by sending test notifications through Firebase Messaging on your FCM with the correct custom data:
```
Custom Data: type, with value station for this type of notification
Custom Data: device_id, with the value of the station's id.
```


pantelisss (commented):
I have an issue when the app is in foreground

pantelisss (approved):
👍

PR #276, opened 2024-08-13
FE-1074 Daily Reward - Show "Split" button and multiple wallet

PavlosTze:
### **Why?**
Show in the Daily Reward screen a reward-splitting button, that will promote a bottom sheet dialog with the wallets included in the reward-splitting process. If the user watching this dialog is a stakeholder of this station then their address should be highlighted. The button should be visible only if there is more than one stakeholder. Also, we need analytics for the button.

### **How?**
- Created `RewardSplit` API Model.
- Created extension function `hasSplitRewards` which is true when `reward_split.size >= 2` and use this function to show/hide the "Show Split" button.
- On `RewardDetailsViewModel`, get if the user is logged-in or not in `init` and when the reward `hasSplitRewards`, pre-fetch the wallet address as we will use it in the "Show Split" button if clicked.
- Wait for the job fetching the wallet address to finish if for any case "Show Split" is clicked before that (which shouldn't be the case as the wallet address is saved on in-memory cache)
- Created the UI for the Bottom Sheet Dialog along with the reusable `list_item_reward_split_stakeholder` which represents a row for each different stakeholder.

#### Dev API
- Check any non-internal station with `reward_split.size <= 1` and make sure "Show Split" button is not visible.
- Check `Agreeable Emerald Solar` station that has `reward_split.size == 2` and make sure "Show Split" button is visible and when clicking it the bottom sheet dialog opens and shows the correct data

#### Local Mock
- In Local Mock, the first row in the "Show Split" dialog should be classified as "Own Wallet" (bold and includes the `(You)` suffix.


pantelisss (approved):
👏

PR #277, opened 2024-08-13
FE-1076 Feature/reward split station settings

PavlosTze:
### **Why?**
Show in Station Settings screen a reward-splitting card, with the wallets included in the reward-splitting process. If the user watching this card is a stakeholder of this station then their address should be highlighted. The Reward Splitting card should be visible only if there is more than one stakeholder.

### **How?**
- Added the layout in both station settings-related XML files.
- In both View Models added the logic of getting the splits-related data with the function `handleRewardSplitInfo` which is responsible for getting the splits and the wallet address (if the user is logged in).

#### Dev API
- Check any non-internal station with `reward_split.size <= 1` and make sure the "Reward Splitting" card is not visible.
- Check `Agreeable Emerald Solar` station that has `reward_split.size == 2` and make sure the "Reward Splitting" card is visible.

#### Local Mock
- In Local Mock, the first row in the the "Reward Splitting" card should be classified as "Own Wallet" (bold and includes the `(You)` suffix.


app/src/main/res/values/strings.xml
```diff
@@ -290,6 +290,7 @@
     <string name="cell_ranking_explanation">Cell ranking is a measure of a station’s performance relative to others within the same cell - a hexagonal area, covering approximately 5 square kilometers (2 square miles) on the planet.\n\nCell capacity is a parameter that is used to define the maximum number of stations that are eligible for rewards in a specific cell. Every cell has a predefined capacity that depends on its geospatial characteristics*.\n\nEvery station is ranked in its cell daily, based on its reward score and its seniority. As long as the station’s ranking is above the capacity threshold, it will be rewarded, whereas getting below it will lead to zero rewards. For example, in a cell with a max capacity of 5 stations, if a station is ranked 3rd it will get rewarded, but if it is ranked 7th, it won’t receive any rewards.\n\nRead more about how our Cell Capacity algorithm works.\n\n* The cell capacity is currently set to the fixed value of 10 rewardable stations, for every cell.</string>
     <string name="show_split">Show Split</string>
     <string name="reward_split">Reward Split</string>
+    <string name="reward_splitting">Reward Splitting</string>
```
pantelisss, line 293: It was by mistake in designs. Now is updated to "Reward Split"

pantelisss (changes requested):
Everything looks good except the title I mentioned.
After this fix, we are good to go

pantelisss (approved):
👏

PR #283, opened 2024-08-20
FE-1083 Remote configurable banner profile screen

PavlosTze:
### **Why?**
Create remote configurable banner for Survey promotions in Profile Screen and make it reusable for future iterations.

### **How?**
- Created a `Survey` data model
- Created `SurveyDataSource`, `SurveyRepository` and `SurveyUseCase` all responsible for fetching the survey-related info from the Firebase Remote Config, construct the `Survey` model and pass it to the `ProfileViewModel`
- Created the reusable component `SurveyCardView` with its `view_survey_card.xml`
- Implemented [Firebase Real-Time updates on Remote Config](https://firebase.google.com/docs/remote-config/get-started?platform=android#add-real-time-listener)

### **Testing**
Play with different cases on the remote config and ensure everything works OK:
1. Lifecycle of the banner (Live-Dismissed-Resurface with new ID)
2. Correct Background, links working etc 
3. Back Button works! 
4. Dismiss banner -> Remote config flag is active -> In the next open of the profile screen the banner is not displayed


app/src/main/java/com/weatherxm/data/services/CacheService.kt
```diff
@@ -307,6 +308,14 @@ class CacheService(
         return preferences.getInt(KEY_DEVICES_OWN, 0)
     }
 
+    fun getLastDismissedSurveyId(): String? {
+        return preferences.getString(KEY_DISMISSED_SURVEY_ID, null)
+    }
+
+    fun setLastDismissedSurveyId(surveyId: String) {
+        preferences.edit().putString(KEY_DISMISSED_SURVEY_ID, surveyId).apply()
+    }
```
pantelisss, line 317: I don't know if it's better to keep a list of the dismissed ids instead of only the last one

pantelisss (changes requested):
Everything looks good, let's discuss about the persisted survey id (single or list)

app/src/main/java/com/weatherxm/data/services/CacheService.kt
PavlosTze: We had discussed it on the refinement Panteli and decided to keep only the last one so the others can be reused in future iterations of the same survey.

pantelisss (approved):

PR #292, opened 2024-09-04
FE-1115 Custom fix in reboot command in Helium by disconnecting automatically

PavlosTze:
### **Why?**
Fixing the rebooting  issue with Helium stations with firmware versions lower than 2.9.0.

### **How?**
Brought back the old code were we send the `ATZ` reboot command and immediately return a success without waiting for an "OK" from the device.

### **Testing**
Test it against a Helium station with lower than 2.9.0 and against one with a higher one (e.g. the latest one).


app/src/main/java/com/weatherxm/data/bluetooth/BluetoothConnectionManager.kt
```diff
@@ -366,4 +368,29 @@ class BluetoothConnectionManager(
             }.collect()
         }
     }
+
+    /**
+     * Custom fix needed for firmware versions < 2.9.0 where ATZ commands
+     * does NOT return an OK before rebooting so we return immediately here
+     */
+    suspend fun reboot(command: String, listener: (Either<Failure, Unit>) -> Unit) {
```
pantelisss, line 376: Since the function name is `reboot`, passing the reboot command as an argument  seems redundant

pantelisss (approved):
Codewise seems ok, except for the comment I've made. 
If you agree, we can change it. Otherwise we are good to go

PR #293, opened 2024-09-04
FE-1116 Notification with device_id is stuck in loading state

PavlosTze:
### **Why?**
Handle device first fetch error in device details in order to close the screen and show an error message.

### **How?**
On the first fetch when we have only the deviceId (which happens only in Notications) we had the following code:
```
deviceDetailsUseCase.getUserOwnedDevice(deviceId).onRight {
    onDeviceAutoRefresh(it)
}.onLeft {
    analytics.trackEventFailure(it.code)
}
```
So I transformed it to:
```
deviceDetailsUseCase.getUserOwnedDevice(deviceId).onRight {
    onDeviceAutoRefresh(it)
}.onLeft {
    /**
     * Fetching the device failed, we should propagate null so that the Activity
     * terminates and not getting stuck in a loading state
     */
    onDeviceFirstFetch.postValue(null)
    analytics.trackEventFailure(it.code)
}
```
And in the Activity we now have:
```
model.onDeviceFirstFetch().observe(this) {
    if (it == null) {
        Timber.d("Could not fetch device")
        toast(R.string.error_device_not_found)
        finish()
    }
}
```

### **Testing**
Send a test notification with a valid `deviceId` and one with invalid and ensure both behave properly.

### DIFF
--- a/app/src/main/java/com/weatherxm/ui/devicedetails/DeviceDetailsActivity.kt
+++ b/app/src/main/java/com/weatherxm/ui/devicedetails/DeviceDetailsActivity.kt
@@ -133,6 +133,13 @@ class DeviceDetailsActivity : BaseActivity() {
             }
         }
 
+        model.onDeviceFirstFetch().observe(this) {
+            if (it == null) {
+                toast(R.string.error_device_not_found)
+                finish()
+            }
+        }
+
         model.onFollowStatus().observe(this) {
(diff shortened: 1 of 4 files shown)


pantelisss (changes requested):
Works fine for invalid device ids and for valid device ids but only for owned/followed devices. Otherwise thows the error

PR #305, opened 2024-09-10
FE-1092 FE-1140 Feature/stations rewards screen

PavlosTze:
### **Why?**
Implementation of new Stations Rewards screen.

### **How?**
- Created the respective `ApiModels` which are needed in order to parse the new responses in the 2 new `get` calls in `ApiService` (which are used to populate this screen)
- Created the `getDevicesRewardsByRange` and `getDeviceRewardsByRange` in DataSource & Repository which respectively use these new API calls in order to get the data
- Created the respective `UIModels` which are used to populate the UI and are passed to the view layer.
- These `UIModels` are created in `RewardsUseCase` and `DevicesRewardsViewModel` using the API models.
- Created `DevicesRewardsActivity` which is the activity of the screen along with its `DevicesRewardsViewModel`
- Created `RangeSelectorView` component which is the "7D/1W/1Y" selector above the charts in order to reuse it anywhere we want.
- Created `TooltipMarkerView` component which is the tooltip in the charts and is currently used in both "Total Earned" and "Rewards Breakdown" chart.
- Created `DeviceRewardsAdapter` along with its view which is the list item for every "Device Reward Summary" card in this screen
- Created `DeviceRewardsBoostAdapter` along with its view which is the list item for every "X boost reward details" table of information
- Created `initTotalEarnedChart` and `initRewardsBreakdownChart` functions in `util/Charts` in order to initialize, populate and show these 2 new charts properly.
- Created `ZonedDateTime.getFormattedMonthDate()` function which returns a `ZonedDateTime` in `DD/MM` or `MM/DD` depending on user locale.
- Created `RetryCardView` which is a reusable card as per the design, which is used in the case of errors in this screen inside each card respectively.

### **Testing**
For now it can be tested only via local mock flavour, the respective files which can be edited are `get_user_device_rewards_summary.json` and `get_user_devices_rewards.json`


PavlosTze:
This is ready for review @pantelisss and includes the latest iteration of the error state.

app/src/main/java/com/weatherxm/ui/common/UIModels.kt
```diff
@@ -502,6 +508,15 @@ data class LineChartData(
 
         return dataSets
     }
+
+    fun getEntryValueForTooltip(position: Float): Float {
+        val value = entries.getOrNull(position.toInt())?.y ?: 0F
+        return if (value.isNaN()) {
```
JohnFrontzos, line 514: you can use also the `.takeIf()` here, something like 

`return entries.getOrNull(position.toInt())?.y?.takeIf { !it.isNaN() } ?: 0F`

JohnFrontzos (approved):
A minor suggestion only. Great job Pavlos 🎉

PR #307, opened 2024-09-11
FE-1127 Feature/forecast details improvements

PavlosTze:
### **Why?**
Station Details - Forecast Details improvements

### **How?**
- Updated `WeatherMeasurementCardView` component which reflects the measurement view in the daily card in the forecast details screen, to have the new UI and added it respective in a new daily card in the forecast details screen
- Created two new colors for the charts `chart_primary_line`, `chart_secondary_line` which we use for the chart lines in the forecast and history charts.

### **Testing**
1. Ensure that the new screen looks OK based on the proposed changes.
2. Make sure the chart coloring updates are reflected in the History screen as well.


pantelisss (changes requested):
My only concern is that it scrolls a little more than expected. eg when I tap the Daily Precip. icon the top bar covers the ghaphs title and legend

PavlosTze:

pantelisss (approved):
👏

PR #310, opened 2024-09-11
FE-1116 Notification with device_id is stuck in loading state

PavlosTze:
### **Why?**
Notification with device_id is stuck in loading state so I had to refactor the way we handle device notification and pass everything through DeepLinkRouter for consistency.

### **How?**
- Refactored `UrlRouterActivity` to `DeepLinkRouterActivity` in order to handle all type of possible deep links, both notifications and urls.
- Added `deviceId` in `WXMRemoteMessage` Data Model to use this model when getting a device notification and handling it with the `DeepLinkRouter`.
- Removed support of opening Device Details screen only with the `deviceId`. Now a valid `Device` object is mandatory.

### **Testing**
Ensure all type of notifications are working properly (announcement & station) in both foreground and background.


pantelisss (approved):
👏

PR #327, opened 2024-09-20
FE-1146 FE-1148 FE-1147 FE-1161 Fixes in Rewards Analytics

PavlosTze:
### **Why?**
1. Wrong bottom padding in total earned card in devices list
2. Unexpected behavior in numerous occasions when fetching data in devices list in that screen

### **How?**
- Fixed the padding in the total earned card
- Refactor the way we fetch data and populate the device adaper in the rewards screen due to numerous errors, we use an explicit `Status` param in order for the adapter to be able to know the state and the `ViewModel` (and only that!!) is responsible for sending each state when new data needs to be fetched.

### **Testing**
Except the issues that are already reported, ensure that all states work correctly in the devices list in this screen such us:
1. Error state
2. Loading state
3. Success state
4. With different than default date range selected
5. With the card expanded or not


app/src/main/java/com/weatherxm/ui/devicesrewards/DeviceRewardsAdapter.kt
```diff
@@ -180,6 +191,15 @@ class DeviceRewardsDiffCallback : DiffUtil.ItemCallback<DeviceTotalRewards>() {
...
+            oldItem.details == newItem.details &&
+            oldItem.details.total == newItem.details.total &&
+            oldItem.details.mode == newItem.details.mode &&
+            oldItem.details.status == newItem.details.status &&
+            oldItem.details.totals.size == newItem.details.totals.size &&
+            oldItem.details.boosts.size == newItem.details.boosts.size &&
+            oldItem.details.datesChartTooltip.size == newItem.details.datesChartTooltip.size &&
+            oldItem.details.baseChartData == newItem.details.baseChartData &&
+            oldItem.details.betaChartData == newItem.details.betaChartData &&
+            oldItem.details.otherChartData == newItem.details.otherChartData
```
pantelisss, line 203: Since we have the first check (`oldItem.details == newItem.details`), are the rest required?

pantelisss (changes requested):
I noticed some issues
- The dotted line indicator in graphs is blue
- $wxm is missing from `Rewards breakdown` card (earned by this station, tokens earned so far, total tokens)
- There are commas in boost period dates
- Tapping anywhere on an expanded station card, collapses the card. Can we limit the interaction in the title view area?
- Can we dismiss the graph toolkit on mode changes (7D, 1M, 1Y)?
- Sometimes, playing with modes (7D, 1M, 1Y), it moves me back to home. I'm not sure if this is a crash that is not detected because the debugger is not attached 
- In the following case seems the Y max doesn't work
Screenshot 2024-09-24 at 16 18 46
- Here there is base reward with zeros, should we show them?
Screenshot 2024-09-24 at 16 32 38

PavlosTze:
Good catches @pantelisss , I fixed all of them except two:
- **Regarding your "crashes":** I couldn't find anything in the logs or reproduce it, if you find a specific pattern please let me know. In order for any crashes to be reported, please try with a `release` flavor (non-debug).
- **Regarding the "base reward with zeros":** Yes that's correct. I think we should not show them. That's why I have added the following code in the tooltip:
```
baseData.isDataValid().takeIf { it }?.also {
    baseTitleView.visible(true)
    baseView.visible(true)
    baseValue = baseData.getEntryValueForTooltip(e.x)
    baseView.text = formatTokens(baseValue)
} ?: run {
    baseTitleView.visible(false)
    baseView.visible(false)
}
```
Where `fun isDataValid() = timestamps.isNotEmpty() && entries.filterNot { it.y.isNaN() }.isNotEmpty()`

In the UseCase we map all zeros to `NaN` in order to have proper gaps in the chart when a X boost starts after a period in a chart. So in our case, when any reward is missing completely, then I don't think there is a reason of showing it to the user as it might confuse people more than it could help them.

pantelisss (approved):
🚀

PR #339, opened 2024-09-24
FE-1153 Feature/analytics for reward analytics

PavlosTze:
### **Why?**
Stations Reward screen - Analytics events

### **How?**
Added the 3 events per the spec.

### **Testing**
Ensure that events are logged properly when needed.


pantelisss (approved):
✌️

PR #346, opened 2024-10-01
FE-1168 Fix colors in tooltip

PavlosTze:
### **Why?**
Fix colors in tooltip in dark mode and light mode.

### **How?**
Use the new colors per the design.

### **Testing**
Ensure the tooltip looks the same as the design in light and dark theme.


pantelisss (approved):
👏

PR #364, opened 2024-10-08
FE-1172 Add a warning for CEX Wallets

PavlosTze:
### **Why?**
Prevent users from adding a CEX wallet address in the app

### **How?**
- Added a new textview containing the new CEX warning
- Replaced the text in the "Wallet Compatibility" card
- Removed some unused resources

### **Testing**
Ensure that the text in the two cards in the My Wallet screen is the same as the design.

### **Additional context**
Removed some unused resources.


pantelisss (approved):
👌

PR #371, opened 2024-10-14
FE-1186 FE-1192 Station Health in Stations List

PavlosTze:
### **Why?**
1. Align the threshold for Reward score error status to be the same as the QoD error.
2. Station List card improvements for health metrics

### **How?**
- Integrated the new `metrics` object in the API response in the `Device` object
- Added the respective `qodScore`, `polReason` and `metricsTimestamp` in the `UIDevice` object in order to use them aftewards for UI purposes
- Redesigned the device card and integrated the new alert chips
- Removed `CellDeviceListAdapter` as it was duplicate to `DevicesAdapter` which we now use in the cell's details
- Created some extension functions in `Views.kt` in order to reuse them for station-health and alerts related UI components

### **Testing**
Ensure that everything looks OK in all different cases based on the specs and the acceptance tests.


app/src/main/java/com/weatherxm/ui/widgets/selectstation/SelectStationAdapter.kt
```diff
@@ -147,6 +165,49 @@ class SelectStationAdapter(private val stationListener: (UIDevice) -> Unit) :
...
+                }
+                binding.issueChip.text =
+                    itemView.context.getString(R.string.issues, item.alerts.size)
+            } else {
+                when (item.alerts[0]) {
+                    DeviceAlert.createWarning(DeviceAlertType.LOW_BATTERY) -> {
+                        binding.issueChip.lowBatteryChip()
+                    }
+                    DeviceAlert.createError(DeviceAlertType.OFFLINE) -> {
+                        binding.issueChip.offlineChip()
```
pantelisss, line 199: Can we encapsulate somewhere this functionality to avoid duplication with the devices' adapter?
In general, in the iOS app, we declare an extension of the UIDevice model (`DeviceDetails` in iOS codebase) which generates this kind of stuff

app/src/main/java/com/weatherxm/util/Rewards.kt
```diff
@@ -6,13 +6,23 @@ import com.weatherxm.ui.common.AnnotationGroupCode
...
 
+    @Suppress("MagicNumber")
+    fun isQodError(score: Int?): Boolean {
+        return score != null && score < 20
+    }
+
+    @Suppress("MagicNumber")
+    fun isQodWarning(score: Int?): Boolean {
+        return score != null && score in 20..79
+    }
```
pantelisss, line 17: My suggestion is to declare a function that returns an error type (warning, error) to keep things cleaner. You can keep these functions with a modified body (eg.  `warningType(score) == .error`)

pantelisss (changes requested):
Great job, just check my comments. To make things clear from my perspective.

PavlosTze:
Good catches @pantelisss , I implemented the optimizations.

pantelisss:
@PavlosTze I've noticed also that in devices with `NO_LOCATION_DATA` the `address` is not visible and shows "No Location" text.
Screenshot 2024-10-17 at 12 31 32

pantelisss (commented):
Looks good!
Let's wait for the location decision before proceeding

pantelisss (changes requested):
@PavlosTze I set pol_reason to null in the mock json and the hexagon color was green

PavlosTze:
@pantelisss The whole `metrics` object should be set to null in order for the "grey" state to appear. Setting just the `pol_reason` to null when the rest of the data are there means that everything is OK and there is no PoL-related error.

pantelisss:
> @pantelisss The whole `metrics` object should be set to null in order for the "grey" state to appear. Setting just the `pol_reason` to null when the rest of the data are there means that everything is OK and there is no PoL-related error.

Sorry, my bad 😞
All good!

pantelisss (approved):

PR #372, opened 2024-10-14
FE-1195 FE-1256 Station Health in Station Details

PavlosTze:
### **Why?**
Station Health bar in Station details screen

### **How?**
- Integrated the new data quality and address cards
- Integrated the new alerts and removed the overflow offline errors
- Updated the top bar

### **Testing**
Ensure that everything looks OK in all different cases based on the specs and the acceptance tests.


app/src/main/java/com/weatherxm/ui/devicedetails/DeviceDetailsActivity.kt
```diff
@@ -208,41 +186,35 @@ class DeviceDetailsActivity : BaseActivity() {
...
+            }
         }
 
-        with(binding.follow) {
+        with(binding.relationBtn) {
             when (device.relation) {
                 DeviceRelation.OWNED -> {
                     setOnClickListener {
-                        // NO-OP
+                        toast(R.string.you_are_owner_of_station)
```
pantelisss, line 199: The toast is presented as expected but the text is invisible

pantelisss (changes requested):
Check my comment. 
One more thing, if we agree that there is "No Location" when pol is `NO_LOCATION_DATA` the user should not be able to navigate to cell station list on location button tap

pantelisss (approved):

PR #380, opened 2024-10-22
FE-1263 My profile: Total earned & Total claimed tokens error state

PavlosTze:
### **Why?**
Use the error state with `N/A` when we encounter an API error in the rewards in the profile screen.

### **How?**
Show the `N/A` values respectively.

### **Testing**
Reproduce it with an error (e.g. turn internet off and go to profile screen) and ensure everything looks according to the spec.


pantelisss (changes requested):
The claim button should be hidden in case of an error. 
Initially, it was visible in the designs, but now they are updated

pantelisss (approved):
👌

PR #381, opened 2024-10-23
FE-1276 Hide loading spinner in devices list at the "Total Earned" button

PavlosTze:
### **Why?**
Remove Total Earned progress icon on Android

### **How?**
Removed the spinner.

### **Testing**
Ensure that everything looks OK, the total earned is not displayed until the data are loaded.


pantelisss (changes requested):
Not sure if this is expected, but in the iOS version, we hide the whole button

https://github.com/user-attachments/assets/c5f0a2d1-9e27-4eae-93b4-436d2c8f402a

PavlosTze:

pantelisss (approved):

PR #387, opened 2024-10-30
FE-1299 Fix the content to share in the station settings screen

PavlosTze:
### **Why?**
There is a bug currently with the content copied when clicking the "Share" button to not follow the spec we have agreed onto.

### **How?**
1. Created a function `toFormattedString` in `DeviceInfoItem` UI Model and
2. Fixed the function `parseDeviceInfoToShare` in `BaseDeviceSettingsViewModel` by using the above function.

### **Testing**
Just ensure that the text to share is correct (content-wise and format-wise) as it should be in the same format as the iOS does.


pantelisss (approved):
👌

PR #390, opened 2024-11-01
FE-1306 CountDownTimer helper class

PavlosTze:
### **Why?**
Create a Count Down helper class to encapsulate `CountDownTimer` in order to reuse this class and be able to unit test the functionality as `CountDownTimer` is an Android library.

### **How?**
- Created a `CountDownTimerHelper` class to encapsulate the `CountDownTimer` functionality and use this class in `BluetoothHeliumViewModel` and `ClaimHeliumPairViewModel`.
- Fixed the unit tests that were missing due to having this Android class directly in our ViewModels.

### **Testing**
1. Ensure that the timer works correctly (can be checked in BLE scanning in the Helium's claiming flow as the `CountDownTimer` is used to show the progress which is shown in the "Scan Again" button)
2. Just make sure the tests run successfully (can be checked through the GitHub Action below as well).

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -370,6 +370,7 @@ tasks.register("jacocoCoverageTestReport", type = JacocoReport::class) {
         "**/ui/widgets/currentweather/*.*",
         "**/ui/Navigator.class",
         "**/util/ChartsKt*.*",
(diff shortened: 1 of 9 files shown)


pantelisss (approved):
👏

PR #406, opened 2024-11-06
FE-1323 Update No Data message for public (non-owned) device

PavlosTze:
### **Why?**
Update No Data message for public (non-owned) device

### **How?**
1. Updated the `no_data_message_public_device` to the new text
2. Created a new `station_inactive_alert_message_public` for the alerts screen as we reused the above one before.

### **Testing**
Ensure both texts look correct on non-owned stations.


pantelisss (changes requested):
Works as expected in devices' list, however in the device details shows the message for owned stations
Screenshot_20241106_165846

PavlosTze:

pantelisss (approved):

PR #410, opened 2024-11-08
FE-1302 Feature/inject dispatchers

PavlosTze:
### **Why?**
As taken from the [Best practices for coroutines in Android](https://developer.android.com/kotlin/coroutines/coroutines-best-practices#inject-dispatchers) and due to some issues wevencountered we need to inject `Dispatchers` where possibly (e.g. in all `ViewModels`) instead of hardcoding them.

### **How?**
1. Inject Dispatchers throughout our app instead of hardcoding them
2. Fix unit tests to use an injected dispatcher in the ViewModels

### **Testing**
1. Run the unit tests and ensure they have 100% success rate.
2. Make a quick iteration throughout the app and ensure no crashes/infinite loadings take place.


pantelisss (approved):
👌

PR #415, opened 2024-11-11
FE-1331 Optimize `isLoggedIn` functionality and the architecture of some usecases

PavlosTze:
### **Why?**
- Optimize the `isLoggedIn` to return a `Boolean` value and 
- Some architecture fixes in some UseCases having wrong/duplicate functions

### **How?**
- Refactored the `isLoggedIn` function in `AuthRepository` to return a `Boolean` instead of an `Either<Failure, Boolean>`.
- Removed the `isLoggedIn` function from all use-cases **except** the `AuthUseCase` and inject this UseCase in all ViewModels that needed it, in order to avoid having duplicate functionality in multiple UseCases.
- Removed some user-related functions from `AuthUseCase` and moved them to `UserUseCase`
- Removed some auth-related functions from `PreferenceUseCase` and moved them to `AuthUseCase`
- The `getAuthToken` and `setAuthToken` which are functions speaking with the `CacheService` are non-suspend functions in order to simplify the code and avoid using `runBlocking` in some cases.

### **Testing**
Test that everything looks OK for the following screens **in both logged-in and logged-out cases**:
1. Login Screen
2. Preferences Screen
3. Select a station for a widget screen
4. Cell Info Screen
5. Device Details Screen

### DIFF
--- a/app/src/main/java/com/weatherxm/data/Modules.kt
+++ b/app/src/main/java/com/weatherxm/data/Modules.kt
@@ -528,7 +528,7 @@ private val usecases = module {
         UserUseCaseImpl(get(), get(), get(), get())
     }
     single<PreferencesUseCase> {
-        PreferencesUseCaseImpl(get(), get(), get())
+        PreferencesUseCaseImpl(get(), get())
     }
     single<DeleteAccountUseCase> {
         DeleteAccountUseCaseImpl(get(), get())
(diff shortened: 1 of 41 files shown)


pantelisss (changes requested):
Everything looks good except one thing I noticed in the widget.
I don't know if it's an emulator issue, the widget doesn't come back to logged in state.
The steps I followed
Logged in -> add a widget -> log out -> the widget state changes as expected -> log in again -> the widget stays in logged out state

PavlosTze:
> Everything looks good except one thing I noticed in the widget. I don't know if it's an emulator issue, the widget doesn't come back to logged in state. The steps I followed Logged in -> add a widget -> log out -> the widget state changes as expected -> log in again -> the widget stays in logged out state

Seems to be an emulator issue, I can't reproduce it in my device. I reproduced it in the emulator but it happens randomly, as there were cases that it worked correctly also. 

But the `isLoggedIn` functionality **is not** related to the widget's UI etc so I think we're good to go regarding this PR.

pantelisss (approved):
So, let's move on!

PR #416, opened 2024-11-14
FE-1335 FE-1336 Fix info button clickable by mistake. Fix percentage layout of hourly forecast

PavlosTze:
### **Why?**
1. Device Observations: Info icon is clickable
2. Device Forecast hourly card: Percentage symbol is missing

### **How?**
1. Replaced in the info button the `ImageButton` with `ImageView`
2. Make in the hourly forecast card in the percentage layout `layout_width="wrap_content"`

### **Testing**
Ensure both issues are fixed.

### DIFF
--- a/app/src/main/res/layout/fragment_device_details_current.xml
+++ b/app/src/main/res/layout/fragment_device_details_current.xml
@@ -203,9 +203,8 @@
                     android:layout_width="match_parent"
                     android:layout_height="wrap_content">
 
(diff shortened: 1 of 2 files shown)


pantelisss (changes requested):
Everything looks good but I've noticed that the title of the first tab in station details is still "Observations" instead of "Overview"

pantelisss (approved):
👌

PR #417, opened 2024-11-15
FE-1068 Optimizations of Weather object (and related unit tests)

PavlosTze:
### **Why?**
The `Weather` object contained very old code that needed a refactor so it can be cleaner, have just one purpose and remove some Android-specific components.

### **How?**
- Created `UnitSelector` object which has a single purpose, to serve the selected (or default) unit for each weather measurement. In all cases we had `Weather.getPreferredUnit` we now use this `UnitSelector`.
- Created a new `data class WeatherUnit(val type: WeatherUnitType, val unit: String)` in order for the `UnitSelector` to return this one, and not make actions based on `Strings` equality.
- Added a `Context` argument in most `Weather`-related functions as the `UnitSelector` which is used there, needs it. It's better code-wise than inject it, as it's UI-related code there and we do not store/reference this context anywhere.
- Removed the `ignoreConversion` argument in some `Weather`-related functions as it was needed just for one case, the charts. Moved this functionality specific to Charts in `ChartsView` component.
- Fixed the related unit tests to cover the new cases and remove the old ones. Also instead of using a new `mockk<CacheService>()` where it's needed, a reusable one has been defined at `TestConfig` and all usecases now use this one, as having multiple ones here and there caused issues.

### **Testing**
Ensure that all weather-related data in-app look OK, experiment with different weather units as well.


pantelisss (approved):
👏

PR #422, opened 2024-11-26
FE-1426 Fixes on Helium's claiming result screen

PavlosTze:
### **Why?**
The UI in the Helium's claiming result screen was outdated and had several inconsistencies with the design.

### **How?**
- Removed the `BleActionFlowView` component from the result screen and added in that fragment only the components and views that were needed in our purpose.
- Removed from the `BleActionFlowView` and its layout everything which was related to helium's claiming which led to several optimizations and deletions of code in other activities such as Reboot, Change Frequency and OTA.
- Adjusted the `ClaimHeliumResultFragment` and its layout in order to use the new UI which is according the design.

### **Testing**
Ensure that the current result screen now looks according to the design in success/error states.


pantelisss (approved):
I couldn't test it on a device. Codewise looks nice

PR #437, opened 2024-12-09
Feature/fe 1344 photo capturing and gallery screens

PavlosTze:
### **Why?**
Implement the screens where users can capture and manage photos before uploading.

### **How?**
- Created `PhotoGalleryActivity` with its ViewModel which is the activity responsible for launching the camera app and showing the photos as the design.
- Added the `UIDevice` object in the intents starting the photo verification flow as we need it for various purposes.
- Moved function `normalizedName()` to `UIDevice`
- Created `StationPhoto` UI Model which holds 2 variables, `remotePath` and `localPath` for each case. In our ViewModel we have a list of `StationPhoto` so we can show, add or delete them.

### **Testing**
Use local mock and start a claiming flow, at the end start the photo verification flow and proceed with testing the gallery + camera.

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -524,6 +524,7 @@ dependencies {
 
     // Image Loader
     implementation(libs.coil.base)
+    implementation(libs.coil.compose)
     implementation(libs.coil.gif)
 
     // Chucker - HTTP Inspector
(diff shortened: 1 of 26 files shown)


pantelisss (changes requested):
Everything looks good! Only two things to discuss 
- I think the newly captured photo should be placed at the beginning of the list and be selected as well
- I think the camera (and permission prompt) should open once the user taps the plus button instead of the activity launch

PavlosTze:
@pantelisss This is ready for testing. Does **not** contain the skipping of terms etc.

pantelisss (approved):
👏

PR #463, opened 2025-01-17
Feature/fe 1345 photo uploading mechanism

PavlosTze:
### **Why?**
Photo Uploading Mechanism

### **How?**
- Created in our flow the functions `deleteDevicePhoto()` (for deleting a photo), `getPhotosMetadataForUpload()` (for fetching a photo's metadata from the API), `getDevicePhotoUploadIds()` (in order to cancel them if needed), `addDevicePhotoUploadIdAndRequest()` (save the request in order to get it again if needed e.g. on retry), `getUploadIdRequest()` (get the request in order to retry it).
- Created  the respective `ApiModels` for the metadata of a photo
- Created the respective API endpoints in `ApiService`
- Fixed UI Model `UploadPhotosState` to better represent the error state
- Created the worker `UploadPhotoWorker` which is responsible of creating an upload request, prepare it and start it.
- Created the `PhotoUploadActivity` and `PhotoUploadViewModel`. In this activity we contact the API to get the metadata, copy the captures photos to cache and copy their EXIF metadata and then start the above worker (**one worker for each photo as we need different, parallel uploads**).
- Created more unit tests that were needed (some last ones are still pending - To Be Completed with the retry functionality).

### **Testing**
Ensure that all tests look OK. Simulator works OK for taking pictures and uploading them etc. You can also emulate bad internet connection by clicking the 3 dots at the top right of the emulator window.

### DIFF
--- a/app/src/local/assets/mock_files/device_photos_list.json
+++ b/app/src/local/assets/mock_files/device_photos_list.json
@@ -1,11 +1,5 @@
 [
-    {
-        "url": "https://i0.wp.com/weatherxm.com/wp-content/uploads/2023/09/5-5.png"
-    },
-    {
-        "url": "https://docs.weatherxm.com/img/wxm-devices/deployments/good-example-1.jpg"
-    },
-    {
-        "url": "https://docs.weatherxm.com/img/wxm-devices/deployments/good-example-2.jpg"
-    }
+    "https://i0.wp.com/weatherxm.com/wp-content/uploads/2023/09/5-5.png",
+    "https://docs.weatherxm.com/img/wxm-devices/deployments/good-example-1.jpg",
+    "https://docs.weatherxm.com/img/wxm-devices/deployments/good-example-2.jpg"
 ]
--- a/app/src/local/assets/mock_files/get_photos_presigned_metadata.json
+++ b/app/src/local/assets/mock_files/get_photos_presigned_metadata.json
@@ -1,0 +1,13 @@
+{
+    "url": "https://weatherxm.com",
+    "fields": {
+        "bucket": "bucket",
+        "X-Amz-Algorithm": "X-Amz-Algorithm",
+        "X-Amz-Credential": "X-Amz-Credential",
+        "X-Amz-Date": "X-Amz-Date",
+        "X-Amz-Security-Token": "X-Amz-Security-Token",
+        "key": "key",
+        "Policy": "Policy",
+        "X-Amz-Signature": "X-Amz-Signature"
+    }
+}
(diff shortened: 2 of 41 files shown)


pantelisss (changes requested):
Great job! 
I've noticed the following issues
- Upload button is always enabled if photos > 2 even if t there are no photos to upload
- I think istruction view with terms toggle should be presented, even if there are stored photos, when terms are not accepted. Currently it's presented if there are no photos
- Photo's card in settings seems it's not updated when you upload the sixth photos (check the video)
https://github.com/user-attachments/assets/ec0b84ce-7ca9-4087-9b78-50544deeb159

- The banner in the home is initially failed before changing to the uploading state (simulated bad network )

- I think banner tap doesn't navigate to device settings when it is in the uploading state
- If, for some reason, there is only one photo, there is no way to go to settings. There are no photos or button to tap in the device settings
- When I tap cancel the UI state in settings is failed. Not sure if has to do with simulated bad network
Screenshot 2025-01-17 at 16 22 35

- Not sure when this happens of if it's a simulator issue
Screenshot 2025-01-17 at 16 13 59

PavlosTze:
Nice findings @pantelisss, thank you!! I fixed your issues, please have a look at it again. More specifically:

> Upload button is always enabled if photos > 2 even if t there are no photos to upload

Fixed.

> I think istruction view with terms toggle should be presented, even if there are stored photos, when terms are not accepted. Currently it's presented if there are no photos

Fixed.

> Photo's card in settings seems it's not updated when you upload the sixth photos (check the video)

Can't reproduce it. Whenever the upload library sends a success, we make an API call to fetch the latest devices, so the only thing I can think of is bad connection there that causes some lag or a failed request? I couldn't reproduce it on the emulator 😞 

> The banner in the home is initially failed before changing to the uploading state (simulated bad network )

Can't reproduce it either. I simulated "Edge" connection, which is a really slow one, but everything looked OK. Didn't start from the failed state. Also, the only way to start from a failed state would be in case of an actual error at one of the uploads (the one before the "In Progress" for example)

> I think banner tap doesn't navigate to device settings when it is in the uploading state

Fixed.

> If, for some reason, there is only one photo, there is no way to go to settings. There are no photos or button to tap in the device settings

Fixed. Although hopefully that won't be possible at all as we need minimum 2 photos.

> When I tap cancel the UI state in settings is failed. Not sure if has to do with simulated bad network

Fixed.

> Not sure when this happens of if it's a simulator issue

Everything! Emulator, compose and slow network issues. No idea why it happens it seems random and not too often. It's one of the reasons I said not to enlarge the icon... I suggest ignoring this for now.

pantelisss (changes requested):
Seems good!
Only one thing I've noticed and we are good to go. When there are photos and the terms are not accepted, the gallery after the instructions view is empty

PavlosTze:
Good catch @pantelisss , fixed, along with the retry functionality as well.

pantelisss (approved):
👏

PR #475, opened 2025-01-23
Legal Terms Prompts and Checks

PavlosTze:
### **Why?**
Integrate the new legal terms as per the spec.

### **How?**
- Created `TermsDialogFragment` which is the terms prompt that is shown in `HomeActivity` and `DeviceDetailsActivity`.
- Created the functions `shouldShowTermsPrompt` and `setAcceptTerms` which get and set the respective values from the cache to the respective activities all the way down with the correct flow.
- Fixed the unit tests to include those 2 functions in all steps.

### **Testing**
Ensure that settings, the registration screen and the 2 dialogs (home screen + device details through widget) work OK.

In order for the device details to test it:
1. Install the app
2. Login
3. Close the app without dismissing the prompt
4. Add a widget
5. Open a widget to get into the device details

### **Additional context**

Closes FE-1571 FE-1570 FE-1555


pantelisss (approved):
👏

PR #480, opened 2025-01-28
Feature/post helium station frequency

PavlosTze:
### **Why?**
Post station's Helium frequency when it's changed.

### **How?**
1. Introduce `DeviceFrequencyDataSource` and the `setDeviceFrequency` endpoint.
2. In `BluetoothConnectionRepositoryImpl` when `setFrequency` is a success then fetch the device EUI and post it through the above datasource, and return the success as `Unit` or the respective `Failure` if any of the functions in the flow fail.
3. 
### **Testing**
Currently not testable as we don't have an API. You can just put a breakpoint at `setFrequency` at `BluetoothConnectionRepositoryImpl` and ensure that the code runs successfully there.

### **Additional context**
Fixes FE-1573, FE-1572


pantelisss (approved):
Couldn't test it on a real device, codewise looks good!

PR #482, opened 2025-01-28
Android new user registration returns API error when first or last name are empty

PavlosTze:
### **Why?**
When first name and last names are empty, replace them with null in sign up.

### **How?**
```
        val first = if (firstName.isNullOrEmpty()) null else firstName
        val last = if (lastName.isNullOrEmpty()) null else lastName
```

### **Testing**
Make two test sign-ups and make sure everything is OK:
1. With first and last name empty
2. With first and last name filled


pantelisss (approved):
👌

PR #484, opened 2025-01-30
Update the relation in the UI when following a station from a cell's list

PavlosTze:
### **Why?**
A **UI** bug that didn't update the relation of a device when clicking "follow" or "unfollow".

### **How?**
In the `DeviceAdapter` also consider `device.relation` to reflect changes in the UI or not.

### **Testing**
Follow & unfollow stations from a cell's list and make sure the UI updates correctly.


pantelisss (approved):
👌

PR #490, opened 2025-02-03
FE-1608 Hotfix/fix crashes

PavlosTze:
### **Why?**
Fix crashes on dialogs (e.g. Terms dialog).

### **How?**
Auto-dismiss the dialog when it's restored from saved instance (e.g. when the activity is destroyed and restarted) and let it be the caller's responsibility to re-instantiate it.

### **Testing**
- On the device go to "Developer Options" -> "Don't keep activities"
- Open a fresh install so the terms can be shown
- Click the home button to let the app go to the background or any link in the dialog
- Re-open the app
- Make sure it does **not** crash.


pantelisss (approved):
Seems good👏

PR #491, opened 2025-02-03
Display helper text when max photos reached

PavlosTze:
### **Why?**
Show a helper text when max photos are reached.

### **How?**
Created the composable `MaxPhotosReached` and show/hide it depending on the current photos number.

### **Testing**
Make sure it looks OK as per the design.


PavlosTze:
@pantelisss Ready, I implemented the snackbar.

pantelisss (approved):
👏

PR #492, opened 2025-02-04
Reset widget to 'Select Station' when clear app data takes place

PavlosTze:
### **Why?**
Widgets fails to load after clearing app data on Android

### **How?**
Created `clearAllWidgets` in the Worker which is responsible for getting the widget IDs and sending "Should Select Station" events to them.

### **Testing**
1. Add a widget
2. Close the app and clear its cache & data
3. Open the app
4. Ensure the widget is at the "Select Station" state
5. Login through the app
6. Select a station through the widget
7. Ensure everything looks OK.

### **Additional context**
Current limitations: Works only for 1 widget currently.


pantelisss (approved):
Looks good!

PR #496, opened 2025-02-06
Empty Device List redesign

PavlosTze:
### **Why?**
Empty Device List redesign.

### **How?**
Redesigned the empty state by making fixes in both `fragment_devices` and `activity_home` and added the unit tests missing as we have a new function in `DevicesViewModel`.

### **Testing**
On an account with no devices (either by mock editing the `get_user_devices.json` or by creating a new account) ensure that the new UI looks OK.


pantelisss (changes requested):
Good job!
I've noticed the following 
- In case the info view is on screen, we have overlaps. Maybe we should reach the design team for this


- I think the main content, everything except the arrow view, should be centered in the highlighted area



- When I change tab (eg. profile) and come back home, the arrow animation is fired again, which, I think, is nice and expected. This doesn't happen if I come back from the claim station screen (+ button)

PavlosTze:
> Good job! I've noticed the following
> 
>     * In case the info view is on screen, we have overlaps. Maybe we should reach the design team for this
> 
> 
> 
> 
>     * I think the main content, everything except the arrow view, should be centered in the highlighted area
> 
> 
> 
> 
>     * When I change tab (eg. profile) and come back home, the arrow animation is fired again, which, I think, is nice and expected. This doesn't happen if I come back from the claim station screen (+ button)

I fixed the last bullet Panteli. The above two, let's wait for the design's team feedback as these are currently in different layouts, so it's not possible to alter their alignment or sth. It needs a refactor of this PR.

PavlosTze:
@pantelisss This is ready for review based on the latest requirements.

app/src/main/java/com/weatherxm/ui/home/HomeActivity.kt
```diff
@@ -176,6 +161,62 @@ class HomeActivity : BaseActivity(), BaseMapFragment.OnMapDebugInfoListener {
...
+            model.getInfoBanner()
+            checkForNoDevices()
+        } else {
+            binding.emptyContainer.visible(false)
+            binding.claimRedDot.visible(false)
+        }
+    }
+
+    private fun checkForNoDevices() {
+        if (devicesViewModel.hasNoDevices()) {
```
pantelisss, line 188: Not sure if we should check for any device or only for owned

pantelisss (approved):
👏

PR #497, opened 2025-02-06
Optimizations in Device Settings Activities

PavlosTze:
### **Why?**
Multiple code duplicates, try having a parent activity that will encapsulate most of the UI-related code in order to centralize it.

### **How?**
Create `BaseDeviceSettingsActivity` in order to capture most of the code that can be reused between the WiFi and Helium activities. Those two activities (`DeviceSettingsWifiActivity` and `DeviceSettingsHeliumActivity`) extend the `BaseDeviceSettingsActivity`.

### **Testing**
Play around with both WiFi and Helium station settings and make sure everything is working properly and nothings has broken.


pantelisss (approved):
Looks good!

PR #506, opened 2025-02-13
Hide info annotation banner on rewards when reward score is >=99%

PavlosTze:
### **Why?**
Hide warning banner when stations has 99% score

### **How?**
Created `shouldShowAnnotations` functions in the respective models and call them before showing any annotations.

### **Testing**
Play with mock `get_device_rewards.json` and `get_device_reward_details.json` and ensure the annotations are not shown when reward or QoD score is 99% or 100%.


pantelisss (approved):
👏

PR #507, opened 2025-02-13
On Delete photo -> wait for API response before deleting it from Gallery

PavlosTze:
### **Why?**
On Delete photo -> wait for API response before deleting it from Gallery

### **How?**
Wait for API before deleting a remote photo. also add linear loader and snackbar in case of error.

### **Testing**
Perform photo deletions and ensure everything looks OK with/without error (e.g. with error = no connection or replace an argument in `usecase.deleteDevicePhoto` in the ViewModel's call place with an invalid one).


pantelisss (approved):
👌

PR #535, opened 2025-03-07
Catch two obvious crashes in OTA and Change Frequency

PavlosTze:
### **Why?**
Catch two obvious crashes in OTA and Change Frequency

### **How?**
1. Catch `IOException` that the `bytes()` function might throw.
2. Do not fire the `onSet` listener when the selected item position is `INVALID_POSITION == -1` (i.e. a frequency NOT selected yet)

### **Testing**
Not sure, I didn't manage to reproduce either of them. You can try to reproduce the second one by commenting out the line 77 in `ChangeFrequencyViewModel`, no idea how a user has actually reproduced it though. I fixed the end result of it.


pantelisss (approved):
Codewise looks good 👍

PR #543, opened 2025-03-19
Create Pro promotional card inside some screens

PavlosTze:
### **Why?**
Create Pro promotional card inside some screens.

### **How?**
Created the reusable composable `ProPromotionCard` and added it in the screens:
- Station Forecast tab
- Forecast Details
- Profile
- Station History
- Cell Station List
- Network stats

### **Testing**
Ensure that in all screens looks OK and according to the design and that on every screen clicking the button opens the respective bottom sheet dialog.


pantelisss (changes requested):
Everything looks good, my only concern is about the clickable area. I think that only the "Get Pro" button should be clickable, not the entire card

PR #567, opened 2025-04-14
Feature/fe 1789 create upload from gallery button

pantelisss:
### **Why?**

Functionality to pick photo from phone gallery 

### **How?**

- Added gallery button
- Implemented gallery launcher (`ActivityResultLauncher<PickVisualMediaRequest>`) to present the photo picker. I couldn't find a way to dynamically limit the max photos to pick, so the picker's limit is 1

### **Testing**
Ensure the gallery picker works as expected

### **Additional context**

fixes fe-1789


app/src/main/java/com/weatherxm/ui/photoverification/gallery/PhotoGalleryActivity.kt
```diff
@@ -234,6 +260,11 @@ class PhotoGalleryActivity : BaseActivity() {
                 )
                 getCameraPermissions()
             }
+
+            binding.galleryBtn.setOnClickListener {
+                // TODO: track event
+                openPhotoPicker()
```
PavlosTze, line 266: No need I think to create a different function, just call directly `navigator.openPhotoPicker(photoPickerLauncher)` as it's the only place we call that.

PavlosTze, line 266: No need I think to create a different function, just call directly `navigator.openPhotoPicker(photoPickerLauncher)` as it's the only place we call that.

PavlosTze (changes requested):
Except my comment below, everything else looks to work OK except one issue, the "single-image pick" didn't work. I can still pick multiple files, end up with >6 images to upload and then get an error. Happens on both the emulator and my Samsung.

PavlosTze (approved):
Good job Panteli! Everything works properly 👏

app/src/main/java/com/weatherxm/ui/photoverification/gallery/PhotoGalleryViewModel.kt
```diff
@@ -34,9 +35,9 @@ class PhotoGalleryViewModel(
 
     fun getLocalPhotosNumber() = photos.filter { it.localPath != null }.size
 
-    fun addPhoto(path: String) {
+    fun addPhoto(path: String, photoSource: PhotoSource?) {
```
PavlosTze, line 38: In both calls of `addPhoto`, `photoSource` isn't `null`, so no need to make it nullable here. Let's remove the `?` until we need it.

app/src/main/java/com/weatherxm/ui/common/UIModels.kt
```diff
@@ -684,6 +688,17 @@ enum class RewardTimelineType {
...
+    CAMERA,
+    GALLERY;
+
+    val exifUserComment: String
+        get() = when (this) {
+            CAMERA -> "wxm-device-photo-camera"
+            GALLERY -> "wxm-device-photo-library"
+        }
+}
+
```
PavlosTze, line 701: This can be simplified to:
```
enum class PhotoSource(val exifUserComment: String) {
    CAMERA("wxm-device-photo-camera"),
    GALLERY("wxm-device-photo-library")
}
```

app/src/main/java/com/weatherxm/ui/photoverification/gallery/PhotoGalleryActivity.kt
```diff
@@ -125,7 +149,7 @@ class PhotoGalleryActivity : BaseActivity() {
                     analytics.trackEventUserAction(
                         AnalyticsService.ParamValue.START_UPLOADING_PHOTOS.paramValue
                     )
-                    navigator.showPhotoUpload(this, model.device, model.getPhotosLocalPaths())
```
PavlosTze, line 128: This function `getPhotosLocalPaths` isn't used anymore, so please delete it from the View Model and the respective unit tests.

PavlosTze (changes requested):

app/src/main/java/com/weatherxm/ui/photoverification/upload/PhotoUploadActivity.kt
```diff
@@ -97,7 +97,10 @@ class PhotoUploadActivity : BaseActivity() {
                 val file = File(cacheDir, fileName)
                 val imageBitmap = BitmapFactory.decodeFile(stationPhoto.localPath)
                 file.copyInputStreamToFile(compressImageFile(imageBitmap))
-                copyExifMetadata(stationPhoto.localPath, file.path)
+                copyExifMetadata(
+                    stationPhoto.localPath, file.path,
+                    stationPhoto.source?.exifUserComment
```
PavlosTze, line 102: Let's have `file.path` in a new line please for formatting issues.

app/src/main/java/com/weatherxm/ui/Navigator.kt
```diff
@@ -525,14 +528,14 @@ class Navigator(private val analytics: AnalyticsWrapper) {
...
-        photos: ArrayList<String>
+        photos: ArrayList<StationPhoto>
     ) {
         context?.let {
             it.startActivity(
                 Intent(it, PhotoUploadActivity::class.java)
                     .addFlags(Intent.FLAG_ACTIVITY_SINGLE_TOP or Intent.FLAG_ACTIVITY_CLEAR_TOP)
                     .putExtra(ARG_DEVICE, device)
-                    .putStringArrayListExtra(ARG_PHOTOS, photos)
+                    .putParcelableArrayListExtra(ARG_PHOTOS, photos)
```
PavlosTze, line 538: Instead of that, let's use the extension function found in `Views.kt` called `putParcelableList` , so that the `photos: ArrayList<StationPhoto>` param, will be `photos: List<StationPhoto>` and the `putParcelableList` will handle everything else by itself.

And you can remove the `ArrayList` from the `ArrayList(model.photos.filter { it.isLocal })` as we will pass a `List` now.

PavlosTze (changes requested):

PavlosTze (approved):
👏

PR #578, opened 2025-04-29
Add station count labels to cells on map

PavlosTze:
### **Why?**
Add station count labels to cells on map

### **How?**
- Created `pointManager` and utilized that in order to add points to the cells at zoom level >= 10.
- Also created function `toPointAnnotationOptions` that gives back all the point annotations that should be added in the map in a correct way, and added those points at the UI Model `ExplorerData` to serve them directly.

### **Testing**
Ensure that everything looks and works OK with no lags.

### DIFF
--- a/app/src/main/java/com/weatherxm/ui/components/BaseMapFragment.kt
+++ b/app/src/main/java/com/weatherxm/ui/components/BaseMapFragment.kt
@@ -10,7 +10,9 @@ import com.mapbox.maps.MapView
 import com.mapbox.maps.MapboxMap
 import com.mapbox.maps.Style
 import com.mapbox.maps.plugin.annotation.annotations
+import com.mapbox.maps.plugin.annotation.generated.PointAnnotationManager
 import com.mapbox.maps.plugin.annotation.generated.PolygonAnnotationManager
+import com.mapbox.maps.plugin.annotation.generated.createPointAnnotationManager
 import com.mapbox.maps.plugin.annotation.generated.createPolygonAnnotationManager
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/weatherxm/ui/components/BaseMapFragment.kt
```diff
@@ -96,6 +99,9 @@ open class BaseMapFragment : BaseFragment() {
 
             with(binding.mapView.annotations) {
                 polygonManager = this.createPolygonAnnotationManager()
+                pointManager = this.createPointAnnotationManager().apply {
+                    textSize = 0.0
+                }
```
pantelisss, line 104: There is an alternative approach to avoid zoom observation and textSizes changes.
Use the following snippet (or something like this).
```
                val config = AnnotationConfig(layerId = "point_layer")
                pointManager = this.createPointAnnotationManager(config).apply {
                    textSize = STATION_COUNT_POINT_TEXT_SIZE
                }
```
and onMapReady implementation of `ExplorerMapFragment` add the following line
```
map.getLayer("point_layer")?.minZoom(10.0)
```

pantelisss (changes requested):
Looks good. However I think the layer id solution is cleaner. If you think so, let's update it. Otherwise we are good to go.

app/src/main/java/com/weatherxm/ui/components/BaseMapFragment.kt
PavlosTze: Great suggestion! Fixed!

pantelisss (approved):
👏

PR #585, opened 2025-05-06
Show cell data quality chip

PavlosTze:
### **Why?**
Show cell data quality chip

### **How?**
- Added in the layout Chip `dataQualityChip` which takes as value the respective `cellDataQuality` from the first `UIDevice` to populate it.
- Fixed the unit tests where that `cellDataQuality` had to be defined.
- Added the respective `LEARN_MORE` event when the info button of that new chip is clicked.

### **Testing**
Ensure that the UI looks OK, you can play with local mock version because dev isn't working currently properly.


app/src/main/java/com/weatherxm/data/models/ApiModels.kt
```diff
@@ -80,13 +80,16 @@ data class PublicDevice(
     val cellCenter: Location?,
     val address: String?,
     val metrics: Metrics?,
+    @Json(name = "avg_data_quality")
```
pantelisss, line 83: The key here is `cellAvgDataQuality`

pantelisss (approved):

PR #629, opened 2025-06-18
Feature/fe 1402 token metrics total wxm allocated card

PavlosTze:
### **Why?**
The Total $WXM Allocated card.

### **How?**
- Implemented the API schema which contains the relevant info
- Integrated a new library for the gauge chart, and added in `TokenMetricsActivity` (and its layout) the respective code for that new card.

### **Testing**
Ensure everything looks OK.


PavlosTze:
> Everything looks good. It would be good If we can do something with this. Otherwise we're good to go Screenshot 2025-06-19 at 21 35 07

Which emulator is that? Couldn't reproduce it and it looks really weird given that it has width to capture, and it doesn't. Can you try again please and let me know?

pantelisss (approved):

PR #654, opened 2025-07-08
Fix multiple $opt_in events in Mixpanel

PavlosTze:
### **Why?**
Fix too many opt-in events tracked, issue happening on Android where in the last month we have 278K events.

### **How?**
We check if the user is opted out with `mixpanelAPI.hasOptedOutTracking()` and only if that's true, we proceed with the `optInTracking`.

### **Testing**
Open and close the app multiple times, on the dev flavor, ensure that in the events in Mixpanel Dev, only one Opt In is being shown.


pantelisss (approved):
👏

PR #655, opened 2025-07-10
Feature/fe 1886 background job to schedule station alert notifications

PavlosTze:
### **Why?**
Background job to schedule station alert notifications

### **How?**
- Created the respective functions `setDeviceNotificationTypeTimestamp` and `getDeviceNotificationTypeTimestamp` in the data source, repo and cache which SET/GET the timestamp of the latest notification on a specific device + type.
- Created the worker `DevicesNotificationsWorker` which is responsible for running in the background, performing the required checks and sending a notification if needed. 
- The entry point for that worker is on the `HomeActivity` and whenever the main switch is enabled in a device's notification screen. That worker has the `KEEP` param which means on every `initAndStart` that is called, if there is already an instance pending to run, it won't be replaced (to avoid recreating instances on multiple start-ups or multiple on/off in the workers).
- If the user has no notification permissions, or logged out, or no notifications enabled on his own devices, then the `stopWorker` function is called.

### **Testing**
Install the app and enable notifications and make sure the worker & notifications work properly based on the requirements we have. For testing purposes you can change the 2 hours interval to 20 minutes, in `class DevicesNotificationsWorker` (change `private val UPDATE_INTERVAL = 2.hours.toJavaDuration()` to `private val UPDATE_INTERVAL = 20.minutes.toJavaDuration()`)

### DIFF
--- a/app/src/main/java/com/weatherxm/app/App.kt
+++ b/app/src/main/java/com/weatherxm/app/App.kt
@@ -3,7 +3,6 @@ package com.weatherxm.app
 import android.app.Application
 import android.app.NotificationChannel
 import android.app.NotificationManager
-import android.content.Context
 import com.weatherxm.BuildConfig
 import com.weatherxm.service.GlobalUploadObserverService
 import net.gotev.uploadservice.UploadServiceConfig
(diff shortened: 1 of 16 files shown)


pantelisss (approved):
Seems that works as expected👌

PR #657, opened 2025-07-11
Feature/fe 1906 add the photo instructions gallery steps in the claiming

PavlosTze:
### **Why?**
Added the photo instructions and gallery in the claiming flow.

### **How?**
- Created duplicate code of the instructions and gallery we had already, with some optimizations/fixes (as we didn't need all of the code that was there) and put them in the three different claiming flows (Helium, Pulse, WiFi).
- Removed the previous entry points in the claiming flow -> SPV and its respective code
- Created `prepopulateCapturedPhotos` function which if there are taken photos in the claiming flow, and the user cancels it and returns anytime aftewards, these photos will be prepopulated in the gallery. The filter to find them is if the prefix starts with the serial number based on user's input (or in case of Helium if it starts with `DeviceType.HELIUM.name`).

### **Testing**
Ensure that everything works OK in the claiming flows.


PavlosTze:
> Look good but I've encountered some issues. Chek the attached video.
> 
>     * The `Accept terms` toggle is always off. Shouldn't we persist the value just like the regular flow?
> 
>     * On first launch of camera, I tapped the back button and there is an empty photo in the scroller.
> 
> 
> Screen_recording_20250715_102317.mp4

I fixed the first bullet. I couldn't reproduce the second one in both an emulator and a physical device. Opening camera and going back doesn't do anything. Can you give it one more try please?

PR #673, opened 2025-07-22
Feature/fe 1921 create a new home tab in bottom navigation support for non

PavlosTze:
# **Why?**
The new home navigation + the new flow for non-authenticated users.

## **How?**
- Created the new navigation item in the bottom bar which is associated with `LocationsFragment`
- Moved the info banner & announcement banner to that `LocationFragment` (the rest is empty in the scope of this PR).
- Now on the first app startup, we go to the Analytics Opt In Screen and afterwards on the Home Screen, so I removed the previous activities/entry points that are not used anymore and their related code.
- Removed the direct launches of Login screen in case of force logout. It's not needed anymore as we support non-authenticated users.
- In Cache we persist the last dismissed and last announcement banner in order not to re-show them in case of logout/login.
- In `HomeViewModel` & `HomeActivity` we get some auth-related information such as if the user is logged in, and accordingly we update the UI. We also use that information to pass it downwards to `DevicesViewModel` and `ProfileViewModel` in order to proceed (or not) with their respective API calls.
- Fixed the unit tests

### **Testing**
Ensure that everything looks OK UI-wise. Also need to test multiple flows with/without authentication and what happens after each one (e.g. after login does everything

### DIFF
--- a/app/src/main/java/com/weatherxm/ui/home/devices/DevicesFragment.kt
+++ b/app/src/main/java/com/weatherxm/ui/home/devices/DevicesFragment.kt
@@ -195,7 +195,11 @@ class DevicesFragment : BaseFragment(), DeviceListener {
                     adapter.notifyDataSetChanged()
                     binding.statusView.visible(false)
                     binding.recycler.visible(true)
+                    binding.myStationsTitle.visible(true)
+                    binding.sortFilterBtn.visible(true)
                 } else {
+                    binding.myStationsTitle.visible(false)
+                    binding.sortFilterBtn.visible(false)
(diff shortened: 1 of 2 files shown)


pantelisss (changes requested):
Good job! I've noticed some issues
- In the profile tab the loader is loading forever in case of logged out state
Screenshot_20250723_170508

- The my stations tab doesn't show the full screen loader on first launch
- After a successful login, it navigates me back to profile tab without refreshing the state. The state is refreshed on login button tap

https://github.com/user-attachments/assets/2bab2cd4-e3e5-469f-b04e-ec6bc4f53df3

PR #681, opened 2025-07-28
Feature/fe 1922 add support for location based forecast

PavlosTze:
### **Why?**
The revamp of the app adding support for location-based weather + forecast.

Fixes FE-1925: `Better naming of data sources`.

This PR also includes #675 which has already been approved and refers to the analytics of this feature.

### **How?**
- First of all, starting with FE-1925, I renamed some of address-related and location-related data sources and repositories, to more correct names so that we can now use the `LocationsDataSource`, `LocationsRepository` and `LocationsUseCase` to suit our needs for this revamp.
- Created the new UI Model `UILocation` along with `LocationWeather` and `LocationsWeather` which will be used to propagate data to the UI layer and move them between navigations.
- The above-mentioned datasource and repository, are created in order to pass data for getting, adding and removing saved locations from the cache (`SharedPreferences`).
- The `LocationsUseCase` contains functions that are responsible for getting, adding and removing saved locations from the cache. Also gets the current weather of a given location, and clears the "in-memory cached forecast" in case of a swipe refresh in the UI.
- Created new UI components and screens such as `WeatherLocationCardView`, `EmptySavedLocationsView`, `JoinNetworkPromoCard`, `LocationsFragment` etc. that are responsible for populating this new tab in the home screen.
- Added support of locations in Forecast Details screen + view model. Now this screen accomodates both locations + devices (based on if `UIDevice.isEmpty()` or `UILocation.isEmpty()` on each case) with minor differences in each case.

### **Testing**
Currently not testable against the API as it's not released yet. But you can test it against local mock flavor (in this flavor by default you are always logged in). Play around a bit with it ensuring that most of the stuff look OK and we will do a final test when the API is ready as well with real data.


pantelisss (changes requested):
In general looks good! I've noticed some issues Not sure if has to do with the mock configuration
- Empty state of graphs in forecast details says "historical"
Screenshot_20250729_134439

- Search results are not hidden once you clear the text box
https://github.com/user-attachments/assets/b9b0694c-8055-4a02-98d4-690efad4901a

- I think search bar should be presented above all. Currently the tab bar is visible and the search is active too in the explorer tab

PR #685, opened 2025-07-31
Different handling of update prompts for Solana users.

PavlosTze:
### **Why?**
As of now, we were using auto-generated version names and version codes for our apps. In this PR a similar mechanism is introduced, but for Solana. Instead of using the current tags we use in our prod app, different tags will be used for Solana, only those starting with the prefix `Solana_` (such as, the already created `Solana_4.10.0` and `Solana_5.0.0`).

This ensures that Solana versions don't need to follow and/or match our production app's one and we don't have to create new releases in our prod app to match the Solana ones.

Also, conditional audiences have been created in the Firebase's Remote Config that target our Solana app.
So, in case of an update prompt is found, redirect the users to the Solana DApp Store.

### **How?**
In the `getVersionGitTags` and `getLastVersionGitTag`, we introduced the variable `isSolana` which handles the regex differently. The default config is still used in all flavors except the Solana one which runs the above functions and sets the respective `versionCode` and `versionNumber` with the `isSolana = true`.

### **Testing**
1. Run the app and ensure in the in-app settings that the version number is 5.0.0
2. Play around with the Firebase Remote Config, increase the `android_app_version_code` **for the Solana audience only** and ensure you get an update prompt.

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -18,13 +18,17 @@ jacoco {
     toolVersion = "0.8.12"
 }
 
-fun getVersionGitTags(printForDebugging: Boolean = false): List<String> {
+fun getVersionGitTags(isSolana: Boolean, printForDebugging: Boolean = false): List<String> {
     if (printForDebugging) {
-        println("Filter and print ordered release tags")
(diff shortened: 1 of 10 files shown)


pantelisss (approved):
Looks good! 👍

PR #687, opened 2025-08-01
Retry mechanism in refreshing the JWT token

PavlosTze:
### **Why?**
Retry mechanism in refreshing the JWT token as an optimization in case of an API error.

### **How?**
Added a retry mechanism via recursion, with 2 retries max, the first one with a delay of 2 seconds, the second one with a delay of 5 seconds. If all fail, return the failure.

### **Testing**
Not sure it's easy to test it. I would install the app, check the logs in Android Studio filtering for "JWT", check for the "expires" and just run the app again after the access token expires and ensure that the refresh takes place correctly.
εικόνα
εικόνα

### DIFF
--- a/app/src/main/java/com/weatherxm/data/network/interceptor/AuthTokenAuthenticator.kt
+++ b/app/src/main/java/com/weatherxm/data/network/interceptor/AuthTokenAuthenticator.kt
@@ -17,12 +17,14 @@ import com.weatherxm.data.path
 import com.weatherxm.service.workers.ForceLogoutWorker
 import kotlinx.coroutines.Deferred
 import kotlinx.coroutines.async
+import kotlinx.coroutines.delay
 import kotlinx.coroutines.runBlocking
 import okhttp3.Authenticator
 import okhttp3.Request
 import okhttp3.Response
 import okhttp3.Route
 import timber.log.Timber
+import kotlin.time.Duration.Companion.seconds
 
 class AuthTokenAuthenticator(
     private val authService: AuthService,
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/weatherxm/data/network/interceptor/AuthTokenAuthenticator.kt
```diff
@@ -76,16 +79,36 @@ class AuthTokenAuthenticator(
...
+                            return@runBlocking Either.Left(it)
                         }
+                    }
                 }
-            }
+                .onRight {
+                    refreshRetries = 0
+                    Timber.d("[${request.path()}] Token refresh success.")
+                }
+        }
```
pantelisss, line 111: I suggest limiting the scope of the `refreshRetries` with something like the following
```
    private fun networkRefresh(request: Request, refreshToken: String, refreshRetries: Int): Either<Failure, AuthToken> {
        return runBlocking {
            authService.refresh(RefreshBody(refreshToken)).mapResponse()
                .onLeft {
                    var refreshRetries = refreshRetries
                    refreshRetries++
                    when (refreshRetries) {
                        1 -> {
                            delay(2.seconds)
                            return@runBlocking networkRefresh(request, refreshToken, refreshRetries)
                        }
                        2 -> {
                            delay(5.seconds)
                            return@runBlocking networkRefresh(request, refreshToken, refreshRetries)
                        }
                        else -> {
                            Timber.d("[${request.path()}] Token refresh failed after 3 retries.")
                            // Exceeded retries; stop retrying and propagate error
                            return@runBlocking Either.Left(it)
                        }
                    }
                }
                .onRight {
                    Timber.d("[${request.path()}] Token refresh success.")
                }
        }
    }

```

and initially call

```
networkRefresh(request, authToken.refresh, 0)
```

pantelisss (approved):

PR #713, opened 2025-08-26
Quests Landing Screen

PavlosTze:
### **Why?**
- The landing screen of the quests
- Firebase Anonymous Authentication
- The Firestore integration for fetching data & writing user & progress.

### **How?**
- Added a new tab (Quests) in bottom navigation
- When a user visits the Quests tab, an anonymous authentication takes place via Firebase (if user has already authenticated before then use it directly). As soon as we authenticate then we have the proper `userId` from Firebase.
- Created the respective `DataModels` and `UIModels` responsible for parsing the Firestore objects and used in the UI.
- Created the `QuestsDataSource` which is responsible:
    - for fetching an existing user or create a new one
    - for fetching a user's progress on the onboarding quest or create a new empty one
    - for fetching the onboarding Quest with its steps
- Created the `QuestsRepository` and `QuestsUseCase` which pass along the above info.
- Created the `QuestsFragment` and `QuestsViewModel` along with some Jetpack Compose components which are used inside the `QuestsFragment`.
- The `QuestsViewModel` is responsible for fetching the user, and then in parallel the onboarding quest (and its steps) and the user's progress in that quest.  If fetching the user fails, the process is terminated with an error otherwise the process continues and an error is shown.
- In the UI Model `QuestOnboardingData` created a custom constructor `constructor(questDataFirestore: QuestWithStepsFirestore, userProgress: QuestUserProgress?)` in order to be able to use our Firestore-related data into one appropriate UI Model.

### **Testing**
Play around the UI along with changing your user's progress in Firestore and make sure everything looks OK.


pantelisss (changes requested):
Looks good
Only two issues
- Once you pull to refresh, the refresh loader doesn't disappear.
- The progress bar looks cropped.
Screenshot 2025-08-27 at 11 51 45

pantelisss (approved):

PR #714, opened 2025-08-27
Feature/fe 1957 onboarding quest screen

PavlosTze:
### **Why?**
The onboarding quest screen.

### **How?**
- Created `QuestOnboardingActivity`, its View Model and its layout. In the `QuestOnboardingViewModel` we get the onboarding quest's data, just as we do in the previous screen so nothing has changed here.
- Then based on that data, we show the respective steps as required.
- When all of the steps have been completed but the quest hasn't, the "All set!" button is shown
- When the quest is completed a disabled "Quest completed!" button is shown

### **Testing**
Ensure everything looks OK while playing with the data in Firestore.


pantelisss (approved):
👏

PR #717, opened 2025-08-29
Environment sensors step

pantelisss:
### **Why?**

The environment sensors step

### **How?**

- Some UI fixes
- Handle the case of `ENABLE_ENVIRONMENT_SENSORS` properly
- Some refactoring and fixes mentioned in the previous PR

### **Testing**

Ensure the sensors step works as expected 

### **Additional context**

fixes fe-1959


PavlosTze:
Also in addition to the above comments @pantelisss , I just remembered, we said with Achilleas to show our dialog there when the user clicks enable. So please use `ActionDialogFragment` we have already and proceed to `updateStepState` on approving it (it has already listeners etc.).

PavlosTze (approved):
👏

PR #725, opened 2025-09-01
Feature/fe 1961 connect wallet step

PavlosTze:
### **Why?**
The connect wallet step.

### **How?**
- Added the Solana Mobile Wallet Adapter library along with a Base58 converter library.
- When the CTA is clicked, use the newly created `connectSolanaWallet()` function, this library is responsible for opening the respective user's wallet via the Solana library, prompt the user to connect their wallet, get their address back and proceed to calling the `setWallet()` function
- The `setWallet` function makes calls up to the DataSource, to update the user's wallet for this `chainId` (in our case `SOL`), if it returns successfully then it proceeds to mark the step as completed.
- If any of the above fails at any point, we show a snackbar error to the user.
- Also added a click effect on step selection that was missing before.

### **Testing**
Download Solflare wallet in the emulator, or any other mainstream Solana wallet, create a wallet and then via our app make sure that the "Connect Wallet" step works OK.

### DIFF
--- a/app/src/main/java/com/weatherxm/usecases/QuestsUseCase.kt
+++ b/app/src/main/java/com/weatherxm/usecases/QuestsUseCase.kt
@@ -38,7 +38,9 @@ class QuestsUseCaseImpl(val repository: QuestsRepository) : QuestsUseCase {
         return repository.fetchUser(userId)
     }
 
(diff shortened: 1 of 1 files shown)


pantelisss (approved):
👏

PR #729, opened 2025-09-03
FE-1974 Fix some crashes & optimizations

PavlosTze:
### **Why?**
Fix 2 crashes by either catching them or using the recommended approach
Also closes FE-1979

### **How?**
1. Use the recommended `BundleCompat..` approach in `parcelableList`.
2. If the image is null do not proceed, we had like 5 crashes in total last 5 months, but catching them doesn't hurt at all.
3. Minor optimization for https://youtrack.jetbrains.com/issue/KT-73255
4. Created a new `toBigDecimalSafe` function to fix some crashes due to malformed data from API
5. Added a check in Forecast Details to fix an empty header in **device** forecast details

### **Testing**
Complete a photo upload photo and ensure everything is OK (i.e no crash).


pantelisss (approved):
Looks good!

PR #730, opened 2025-09-03
Feature/fe 1976 timestamps in firestore

PavlosTze:
### **Why?**
Set the proper timestamps of `createdAt`, `updatedAt` and `completedAt` on Firestore.

### **How?**
- Whenever a `SET` action takes place, we use `mapOf` to create anything needed. We can't use `FieldValue` objects directly to our `DataModels` as there is a bug in deserialization.
- Replaced `UPDATE` actions with `updateWithTimestamp` which in addition to updating a specific value, updates the `updatedAt` property as well.
- Instead of hard-coded strings I created some `const val` that we will use for consistency.
- When a quest is completed along with updating the `isCompleted`, we now update the `updatedAt` and set the `completedAt`

### **Testing**
Play around with a quest and ensure all timestamps are saved correctly.


pantelisss (approved):
👌

PR #740, opened 2025-09-10
Handle error codes 429 + force refresh in the profile screen

PavlosTze:
### **Why?**
Handle error codes 429 + force refresh in the profile screen

### **How?**
- Added a new check in `mapResponse` if the code returned is 429 to map it to a `TooManyRequestsError` in order then to handle it respectively and map it to a specific error message.
- Added the correct params to force refresh on the profile screen when a pull-to-refresh takes place or a user connects their wallet and goes back to that screen

### **Testing**
Not suer you can test the error, just make sure that the API calls take place in the correct order I guess via the logcat in Android Studio.


pantelisss (approved):
Codewise looks good

PR #753, opened 2025-09-19
Keep only the "en" locale

PavlosTze:
### **Why?**
Remove unsupported locales (all except en) from apk using resConfig. This change is required for Solana, but should be in all flavors, since we don't have any actual translations.

### **How?**
Following the guide [here](https://developer.android.com/reference/tools/gradle-api/8.8/com/android/build/api/dsl/ApplicationAndroidResources#localeFilters()) I specified only the `en` locale.

### **Testing**
Not sure. Maybe @eeVoskos you can mock-test a new release or sth and see what the configuration files say?


eeVoskos (approved):
I can't make it work. I always get `--_--` as default locale in aapt. Let's merge and re-evaluate in the next submission.

